PR #6, opened 2019-04-08
Add Makefile

jmesnil:
The Makefile will install the operator-sdk by building it if it was not
installed (see
https://github.com/operator-framework/operator-sdk#quick-start)


Makefile
```diff
@@ -0,0 +1,47 @@
+DOCKER_REPO ?= quay.io/
+IMAGE ?= jmesnil/wildfly-operator
```
jmesnil: I've not yet move the image to the quay.io/wildfly repository

PR #11, opened 2019-04-17
Clustering

jmesnil:
* Update container command to bind WildFly public address to the pod's
  hostname
* Fix spec's Env check
* Add clustering example


jmesnil:
@pferraro Could you please check if the [tcp stack for JGroups](https://github.com/wildfly/wildfly-operator/pull/11/files#diff-90b754c0a3256e4407ed2a7bd05f3048R409) looks correct for OpenShift?

Issue #4, opened 2019-04-08
Add annotations for Kubernetes & OpenShift


jmesnil:
following the https://github.com/gorkem/app-labels/blob/master/labels-annotation-for-openshift.adoc  and https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/ guidelines

jmesnil:
fixed by #18

PR #18, opened 2019-05-21
[#4] Use standard annotations

jmesnil:
Annotates resources created by the WildFly Operator with the annotations
recommended by
https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/
&
https://github.com/gorkem/app-labels/blob/master/labels-annotation-for-openshift.adoc

Issue: https://github.com/wildfly/wildfly-operator/issues/4


jmesnil:
This fixes #4

PR #39, opened 2019-07-08
fix typo

luck3y:


luck3y:
@jmesnil I noticed this on the go report. some low hanging fruit :)

PR #43, opened 2019-07-10
Add spec sessionAffinity to allow requests from same client ip are pa…

jimma:
…ssed to same wildfly instance


deploy/crds/wildfly_v1alpha1_wildflyserver_crd.yaml
```diff
@@ -35,6 +35,9 @@ spec:
               type: string
             disableHTTPRoute:
               type: boolean
+            sessionAffinity: 
+              description: If connections from the same client ip are passed to the same WildFlyServer instance/pod each time
```
jmesnil: If you add a comment for the `SessionAffinity` field in wildflyserver_types.go, the description will be filled here automatically

jimma: Done.

jmesnil:
@jimma thanks for the contribution!

Issue #51, opened 2019-07-22
createLivenessProbe() and createReadinessProbe() actually reads opeartor container env 


jimma:
If I understand correctly, createReadinessProbe() and createReadinessProbe() reads the operator container Env:
https://github.com/wildfly/wildfly-operator/blob/master/pkg/controller/wildflyserver/wildflyserver_controller.go#L529
https://github.com/wildfly/wildfly-operator/blob/master/pkg/controller/wildflyserver/wildflyserver_controller.go#L557
Should this be configured and read from env from WildflyServerSpec ?

jmesnil:
No, this is not meant to be something configured by the user in theWildFlyServerSpec. 
This is considered an implementation detail internal to the operator.

The reason for these 2 env vars is that EAP has existing CLI scripts that are used as liveness/readiness probes (they are available in the [os-eap-probes modules](https://github.com/jboss-container-images/jboss-eap-modules/tree/master/os-eap-probes/2.0).

The WildFly operator will use the HTTP `/health` endpoint for its liveness (and has no readiness probes).
But the EAP operator will uses these 2 env vars to use its own scripts probes instead by specifying their paths in its own `Dockerfile`.

Does that make sense?

jimma:
This function is executed in wildfly-opeartor container:
`func createLivenessProbe(w *wildflyv1alpha1.WildFlyServer) *corev1.Probe `
 
Does this line only read the wildfly-operator container env before the eap container starts , and the env configured in `DockerFile` for WFLY/EAP image doesn't take effect ?
`readinessProbeScript, defined := os.LookupEnv("SERVER_READINESS_SCRIPT")
`

jmesnil:
The operator code is executed inside the *operator image*.
It reads the env from the *operator image*, not from the *application image*.

It does not have access to any env vars from the application image (whether they are set in its Dockerfile or in the WildFlyServerSpec).

jimma:
So env var  SERVER_READINESS_SCRIPT/SERVER_LIVENESS_SCRIPT are configured through https://github.com/wildfly/wildfly-operator/blob/master/deploy/operator.yaml or  operator image, right ?

jmesnil:
Right. for the EAP operator, I'll likely put the envs in its `Dockerfile` (using Cekit module) so that the operator.yaml does not have to change between WildFly and EAP.

jimma:
I am thinking if the configuration items for liveness and readiness should belong to WildflyServer crds. 
[...]

PR #54, opened 2019-08-05
[#61] introduction of scaledown transaction recovery

### LINKED ISSUE #61: Introduce scaledown transaction recovery
At time the WildFlyServer operator is commanded to scale down the number of pods (aka. the cluster `spec.size` is changed to the smaller number than it was before) the WildFly server is going to be shutdown and the pod is erased. The WildFly could be managing a number of unfinished transactions (some transactions are in in-doubt state) and such transactions would not be never finished.

It's needed to care of in-doubt transactions to be finished before the WildflyServer pod is removed from the cluster for good.

ochaloup:
fixes #61 

Proposal for the scale down processing for transaction recovery.
There are few issues that I'm aware of and that needs to be handled but they could be hopefully handled even after this basic functionality will be merged.

The processing is meant to be run in following steps
* when WildflyServer CR decrease number of pods to be deployed as part of the WildFly cluster the operator start checking if they are clean from in-doubt transactions
* pods are removed from the service to not being hit by a new requests (the label is removed and service does not load balance request to them)
* wildfly management calls are run against the pods to check if there are no in-doubt transactions in the transaction log. Plus verification of WFTC data is cleaned.
* when pods are ready to go the operator let the statefulset definition to be updated and pods are gone

Missing pieces
* <s>when CR is deleted as a whole the recovery processing is not launched. I'm currently not sure how to work with the request on deletion of the whole CR. It seems when the CR is deleted the reconcile loop is informed just afterwards and all child resources are already removed. I need to do investigation on this regards.</s>
*<s> http management calls are run with REST HTTP calls. I didn't find any golang native http digest functionality thus I borrowed some at web. This needs to be confirmed if it's possible.</s>
* <s>tests are needed to be implemented</s>

After discussion with @pkremens and @mnovak adding few more things
* scheduled timeout for EJB beans need to be disabled during scaled down processing. The app server needs to be in graceful shutdown state (possibly, needed to be tested)
* <s>the current functionality missed the piece of the prepared resources without a record in the transaction object store. This check needs to be added to the functionality.</s>
* <s>when transaction is moved to heuristics the scaledown processing could be blocked. A forced scaledown timeout should be defined. After elapsing it the pod needs to be removed even with in-doubt transactions.</s>

For discussion with Jeff
* <s>headless service vs. loadbalancer - how to deal with that</s>

WFLY issue: https://issues.jboss.org/browse/CLOUD-2262


deploy/operator.yaml
```diff
@@ -15,7 +15,7 @@ spec:
       serviceAccountName: wildfly-operator
       containers:
         - name: wildfly-operator
-          image: quay.io/wildfly/wildfly-operator:latest
+          image: 172.30.1.1:5000/wildfly-operator-testing/wildfly-operator
```
pkremens: I believe you've missed this one.

ochaloup: +1, thanks. that was a bit hasty. The change is in my git staging area :man_facepalming:

pkg/controller/wildflyserver/wildflyserver_controller.go
```diff
@@ -130,12 +170,161 @@ func (r *ReconcileWildFlyServer) Reconcile(request reconcile.Request) (reconcile
...
+
+			reqLogger.Info("Going for scaledown with pod", "pod name", scaleDownPod.ObjectMeta.Name, "IP address", scaleDownPodIP)
+
+			managementURL := fmt.Sprintf("http://%v:%v/management", scaleDownPodIP, httpManagementPort)
+			username := string(foundTxnRecoverySecret.Data["username"])
+			password := string(foundTxnRecoverySecret.Data["password"])
+			httpDigest := wildflyutil.HTTPDigest{URL: managementURL, Username: username, Password: password}
+
+			// Enabling recovery listener to speed up recovery processing
+			jsonResult, err := wildflyutil.ExecuteMgmtOp(&httpDigest, wildflyutil.MgmtOpTxnEnableRecoveryListener)
```
jmesnil: could you make a CLI exec instead?

pkg/controller/wildflyserver/wildflyserver_controller.go
```diff
@@ -130,12 +170,161 @@ func (r *ReconcileWildFlyServer) Reconcile(request reconcile.Request) (reconcile
...
+			if err != nil {
+				reqLogger.Error(err, "Cannot enable transaction recovery listener", "Management URL", managementURL, "Pod name", scaleDownPod)
+				scaleDownErrors = append(scaleDownErrors, err)
+				break
+
+			}
+			if !wildflyutil.IsMgmtOutcomeSuccesful(jsonResult) {
+				reqLogger.Info("Failed to enable transaction recovery listener. Scaledown processing can take longer",
+					"Management command", wildflyutil.MgmtOpTxnEnableRecoveryListener, "JSON response", jsonResult)
```
jmesnil: Is there a missing `break` if the op is not successful?

pkg/controller/wildflyserver/wildflyserver_controller.go
```diff
@@ -130,12 +170,161 @@ func (r *ReconcileWildFlyServer) Reconcile(request reconcile.Request) (reconcile
...
+			isReloadRequired := wildflyutil.ReadJSONDataByIndex(jsonResult["response-headers"], "operation-requires-reload")
+			if isReloadRequiredBool, _ := strconv.ParseBool(isReloadRequired); isReloadRequiredBool {
+				jsonResult, err = wildflyutil.ExecuteMgmtOp(&httpDigest, wildflyutil.MgmtOpReload)
+				if err != nil {
+					reqLogger.Error(err, "Cannot reload application container", "Management URL", managementURL, "Pod name", scaleDownPod)
+					scaleDownErrors = append(scaleDownErrors, err)
+					break
+				}
+				for serverStateCheckCounter := 0; err != nil && serverStateCheckCounter < reloadRetryCount; serverStateCheckCounter++ {
+					jsonResult, err = wildflyutil.ExecuteMgmtOp(&httpDigest, wildflyutil.MgmtOpServerStateRead)
```
jmesnil: The operator is blocking until the app server's pod is reloaded, right?

That'd not be correct, the reconcile loop should not be blocking like this.
You would have to issue a :reload and then return `reconcile.Result{Requeue: true}, nil` so that the reconcile loop can handle other `WildFlyServer` events
Once you go back to your own `WildFlyServer` resource, you know from the pod state that it was reloading and then the code can proceed further

pkg/controller/wildflyserver/wildflyserver_controller.go
```diff
@@ -130,12 +170,161 @@ func (r *ReconcileWildFlyServer) Reconcile(request reconcile.Request) (reconcile
...
+				break
+			}
+			if commandResult != "" {
+				reqLogger.Info("Recovery scan to be invoked because of the WFTC data dir is not empty.",
+					"Wildfly Transacton Client data dir path", standaloneServerDataDirPath+"/"+wftcDataDirName, "Output listing", commandResult)
+				isForScan = true
+			}
+			// WFLY still manages in-flight transactions, let's force the recovery scan
+			if isForScan {
+				wildflyutil.SocketConnect(scaleDownPodIP, recoveryPort, txnRecoveryScanCommand)
```
jmesnil: Is there a management op to force recovery scan?

pkg/controller/wildflyserver/wildflyserver_controller.go
jmesnil: or appending to `scaleDownErrors`?

pkg/controller/wildflyserver/wildflyserver_controller.go
```diff
@@ -319,7 +546,9 @@ func (r *ReconcileWildFlyServer) statefulSetForWildFly(w *wildflyv1alpha1.WildFl
 			Labels:    ls,
 		},
 		Spec: appsv1.StatefulSetSpec{
-			Replicas: &replicas,
+			Replicas:            &replicas,
+			ServiceName:         loadBalancerServiceName(w),
+			PodManagementPolicy: appsv1.ParallelPodManagement, // TODO: is it fine to change for Parralel?
```
jmesnil: I'm not sure.
One of the feature of the stateful set is to start one pod after the other. This simplifies clustering code and I'm not sure it should be changed. This can be explored and tested in a future PR

pkg/controller/wildflyserver/wildflyserver_controller.go
```diff
@@ -302,12 +525,16 @@ func matches(container *v1.Container, envVar corev1.EnvVar) bool {
...
 // statefulSetForWildFly returns a wildfly StatefulSet object
-func (r *ReconcileWildFlyServer) statefulSetForWildFly(w *wildflyv1alpha1.WildFlyServer) *appsv1.StatefulSet {
+func (r *ReconcileWildFlyServer) statefulSetForWildFly(w *wildflyv1alpha1.WildFlyServer, txnRecoverySecret *corev1.Secret) *appsv1.StatefulSet {
 	ls := labelsForWildFly(w)
+	delete(ls, labelWildflyOperatorInService)
 	replicas := w.Spec.Size
 	applicationImage := w.Spec.ApplicationImage
 	volumeName := w.Name + "-volume"
 
+	mgmtUser := string(txnRecoverySecret.Data["username"])
```
jmesnil: maybe we should pass directly the user / password credentials in parameter instead of extracting them from the secret?

pkg/controller/wildflyserver/wildflyserver_controller.go
```diff
@@ -447,20 +687,17 @@ func (r *ReconcileWildFlyServer) statefulSetForWildFly(w *wildflyv1alpha1.WildFl
...
 			Namespace: w.Namespace,
 			Labels:    labels,
 		},
 		Spec: corev1.ServiceSpec{
-			Type:            corev1.ServiceTypeLoadBalancer,
-			Selector:        labels,
-			SessionAffinity: sessionAffinity,
+			Type:     corev1.ServiceTypeClusterIP,
+			Selector: labels,
+			// SessionAffinity: sessionAffinity,
```
jmesnil: Why are you commenting `sessionAffinity`?

pkg/controller/wildflyserver/wildflyserver_controller.go
```diff
@@ -447,20 +687,17 @@ func (r *ReconcileWildFlyServer) statefulSetForWildFly(w *wildflyv1alpha1.WildFl
...
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      loadBalancerServiceName(w),
 			Namespace: w.Namespace,
 			Labels:    labels,
 		},
 		Spec: corev1.ServiceSpec{
-			Type:            corev1.ServiceTypeLoadBalancer,
-			Selector:        labels,
-			SessionAffinity: sessionAffinity,
+			Type:     corev1.ServiceTypeClusterIP,
```
jmesnil: Why are you changing the service type?

pkg/controller/wildflyserver/wildflyserver_controller.go
```diff
@@ -504,20 +741,79 @@ func (r *ReconcileWildFlyServer) routeForWildFly(w *wildflyv1alpha1.WildFlyServe
 	return route
 }
 
+// newSecretForCR returns an empty secret for holding the secrets merge
```
jmesnil: s/newSecretForCR/txnRecoveryJBossCliPasswordSecret

pkg/controller/wildflyserver/wildflyserver_controller.go
```diff
@@ -504,20 +741,79 @@ func (r *ReconcileWildFlyServer) routeForWildFly(w *wildflyv1alpha1.WildFlyServe
...
+			"password": password,
+		},
+	}
+	// Set WildFlyServer instance as the owner and controller
+	controllerutil.SetControllerReference(w, secret, r.scheme)
+	return secret
+}
+
+func getWildflyServerPodStatusByName(w *wildflyv1alpha1.WildFlyServer, podName string) *wildflyv1alpha1.PodStatus {
+	i := sort.Search(len(w.Status.Pods), func(i int) bool { return podName == w.Status.Pods[i].Name })
```
jmesnil: why don't you just iterate on w.Status.Pods? afaict you don't need to search in the sorted order.

pkg/controller/wildflyserver/wildflyserver_controller.go
```diff
@@ -33,16 +37,17 @@ import (
...
-	httpApplicationPort         int32 = 8080
-	httpManagementPort          int32 = 9990
-	standaloneServerDataDirPath       = "/wildfly/standalone/data"
+	httpApplicationPort           int32 = 8080
+	httpManagementPort            int32 = 9990
+	recoveryPort                  int32 = 4712
+	standaloneServerDataDirPath         = "/wildfly/standalone/data"       // data directory where runtime data is saved
+	wftcDataDirName                     = "ejb-xa-recovery"                // data directory where WFTC stores transaction runtime data
+	wflyMgmtTxnRecoveryUserName         = "transaction.recovery.scaledown" // username for recovery to run mgmt operations
+	labelWildflyOperatorInService       = "wildfly.operator.in.service"    // label used to remove a pod from receiving load from service during transaction recovery
```
jmesnil: We should follow other labels convention with something like `wildfly.org/service-state` 
The label is stamped on the pod, I'm not sure that mentioning the operator in the name is meaningful..

pkg/controller/wildflyserver/wildflyserver_controller.go
```diff
@@ -284,6 +467,46 @@ func checkUpdate(spec *wildflyv1alpha1.WildFlyServerSpec, statefuleSet *appsv1.S
 	return update
 }
 
+// listing pods which belongs to the WildFly server
+//   the pods are differentiated based on the selectors
+func (r *ReconcileWildFlyServer) getPodsForWildFly(w *wildflyv1alpha1.WildFlyServer) (*corev1.PodList, error) {
+	podList := &corev1.PodList{}
+	labelsForWildfly := labelsForWildFly(w)
+	delete(labelsForWildfly, labelWildflyOperatorInService)
```
jmesnil: would it be simpler to put the labelWildflyOperatorInService when the Pod is actually active instead of deleting it when it should not be queried?

I envisioned `labelsForWildFly`as "static" labels that did not depend on the state of the cluster but it'd not be true anymore.

pkg/apis/wildfly/v1alpha1/zz_generated.openapi.go
```diff
@@ -11,12 +11,12 @@ import (
 
 func GetOpenAPIDefinitions(ref common.ReferenceCallback) map[string]common.OpenAPIDefinition {
 	return map[string]common.OpenAPIDefinition{
-		"./pkg/apis/wildfly/v1alpha1.PodStatus":               schema_pkg_apis_wildfly_v1alpha1_PodStatus(ref),
-		"./pkg/apis/wildfly/v1alpha1.StandaloneConfigMapSpec": schema_pkg_apis_wildfly_v1alpha1_StandaloneConfigMapSpec(ref),
-		"./pkg/apis/wildfly/v1alpha1.StorageSpec":             schema_pkg_apis_wildfly_v1alpha1_StorageSpec(ref),
-		"./pkg/apis/wildfly/v1alpha1.WildFlyServer":           schema_pkg_apis_wildfly_v1alpha1_WildFlyServer(ref),
-		"./pkg/apis/wildfly/v1alpha1.WildFlyServerSpec":       schema_pkg_apis_wildfly_v1alpha1_WildFlyServerSpec(ref),
-		"./pkg/apis/wildfly/v1alpha1.WildFlyServerStatus":     schema_pkg_apis_wildfly_v1alpha1_WildFlyServerStatus(ref),
+		"github.com/wildfly/wildfly-operator/pkg/apis/wildfly/v1alpha1.PodStatus":               schema_pkg_apis_wildfly_v1alpha1_PodStatus(ref),
```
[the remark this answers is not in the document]
ochaloup: strange enough... this is the generated code and I haven't touched that manually. Could be because of a differnt version of 'operator-sdk'? Or maybe I have wrongly set the `GOPATH` or something.

pkg/controller/util/wildfly_mgmt.go
```diff
@@ -0,0 +1,129 @@
...
+//   it expects on arguments to get the response from the HTTP base operation (res)
+//   and the body returned from the HTTP operation that is represented as a JSON
+func IsMgmtOutcomeSuccesful(jsonBody map[string]interface{}) bool {
+	return jsonBody["outcome"] == "success"
+}
+
+// ExecuteMgmtOp executes WildFly managemnt operation represented as a JSON
+//  the execution runs over the HTTPDigest struct
+//  returns the JSON as the return value from the HTTP
+func ExecuteMgmtOp(httpDigest *HTTPDigest, mgmtOpJSON map[string]interface{}) (map[string]interface{}, error) {
```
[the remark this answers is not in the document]
ochaloup: the `CLI` scripts are not needed in this version. My intention was to start using the WFLY HTTP management interface as I think it's a cleaner way to access the management operation. The `ExecRemote` is then more like passing a concatenated strings to some bash operation. Second thing why I think it could be better is that it does not depend on the underlying operating system. If there would be e.g. system without shell script or Windows OS then access via wfly management is independent on that.

This all comes from my strong feeling that operator should be independent on the s2i images. It should cooperate with them smoothly, it should be tested with them but it should not be required to use them for operator to work.

But still there is need the `ExecRemote` call. And the reason is that there is a need to check the content of the data directory of the wfly transaction client. There is no management operation to get information if all remote transactions from the library were processed. Thus in this version of the operator the remote call is needed.

pkg/controller/wildflyserver/wildflyserver_controller.go
```diff
@@ -116,12 +121,47 @@ func (r *ReconcileWildFlyServer) Reconcile(request reconcile.Request) (reconcile
...
+		statefulSetDoesNotExists = true
+	} else if err != nil {
+		reqLogger.Error(err, "Failed to get StatefulSet.")
+		return reconcile.Result{}, err
+	} else {
+		statefulSetDoesNotExists = false
+	}
+
+	// Define secret jboss cli connection for txn recovery processing
+	foundTxnRecoverySecret := &v1.Secret{}
```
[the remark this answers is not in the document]
ochaloup: The `Secrete` is created before the `StatefulSet` creation. But all this logic comes from the way how the reconcile is expected to work. The `StatefulSet` needs the secret before it's created. If secret exists then it could be used. If it does not exist then the secret needs to be created - time for reconcile loop to be requeued. If there is existing `StatefulSet` but somebody removed the `Secret` then the new `Secret` needs to be created and the `StatefulSet` needs to be recreated as it links the non-existent `Secret` at the time.

pkg/controller/wildflyserver/wildflyserver_controller.go
ochaloup: I could but I thought the remote management operation decouples the communication better from the underlaying container. If you think it's better/nicer/whatever I can change all the calls to use the `CLI exec` instead. Just please let me know what is your preference and I can do it.

pkg/controller/wildflyserver/wildflyserver_controller.go
ochaloup: My idea was that if transaction recovery listener was not successfully set then the recovery will be still processing but slowly. Every 2 minutes the recovery will be kicked off. And thus there is chance that after 2 minutes the transactions will be processed.

I do code changes based on the reproaches from the discussion with QE guys and the logic needs to be adjusted and that version will put the error on this operation under `scaleDownErrors`.

pkg/controller/wildflyserver/wildflyserver_controller.go
ochaloup: Yeap, this is one thing I wanted to talk about with you.

For the ejb remote communication and the transaction recovery works correctly the service needs to be headless (https://kubernetes.io/docs/concepts/services-networking/service/#headless-services)
The reason for that is that headless service does not process loadbalancing. The headless service does only DNS resolution. The loadbalancing is the managed by EJB client. The reason for that is that JEB connection and recovery needs the stable connection endpoint. It can't be loadbalanced to arbitrary node in the cluster. The recovery re-connection has to end up at the same node where EJB bussiness logic was processed.

But as I'm writing this I found that I didn't realize that this change influences the other connection processing (e.g. HTTP) which benefits from the load balancing functionality. I think the solution could be to create a new service which would be headless and which would be used for server to server ejb communication. The other service (the original one) would be used for access of other protocols. I need to verify if it's feasible to do so but I think it should work.
What do you think?

pkg/controller/wildflyserver/wildflyserver_controller.go
ochaloup: that was a safeguard ;-) I was not able to find in documentation if the pods are sorted by name in the list. If they are possibly in an arbitrary order then it's dangerous to do so. The statefulset is prescribed to remove the pods in opposite order they were created
```quote
When Pods are being deleted, they are terminated in reverse order, from {N-1..0}
```
(https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#deployment-and-scaling-guarantees)

The recovery processing needs to keep the same ordering for not processing wrong pod than the one the `StatefulSet` removes later.

pkg/controller/wildflyserver/wildflyserver_controller.go
ochaloup: sure, I see. Could be the name be like `wildfly.org/operated-by-service` or similar?

pkg/controller/wildflyserver/wildflyserver_controller.go
ochaloup: yeap, ok, +1

ochaloup:
I pushed the changes in recovery processing to consider not available remote XA resources.

Next week I need to concentrate on tests.

pkg/controller/wildflyserver/wildflyserver_controller.go
ochaloup: I changed this.

pkg/controller/util/wildfly_mgmt.go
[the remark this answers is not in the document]
jmesnil: I don't understand your concern with the scripts. They are running in the container which is Linux, the script will always reliably be executed. 

S2I *is* a requirement from the operator as most of its features will actually be implemented by S2I.

pkg/controller/wildflyserver/wildflyserver_controller.go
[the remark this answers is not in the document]
jmesnil: For code maintenance, it would  be simpler to keep the reconcile loop as linear as possible

1. check if the secret exists
   1.  if it does not exist, create it and reconcile
1. check if the StatefulSet exists
   1. if it does not exist, create it (with the info from the existing secret) and reconcile
   1. if it exists, check that its spec is consistent with the existing secret
      1. if that's not the case, delete the statefulset and reconcile
1. etc....

pkg/controller/wildflyserver/wildflyserver_controller.go
jmesnil: Server-to-server EJB communication are going through HTTP upgrade, right?

pkg/controller/wildflyserver/wildflyserver_controller.go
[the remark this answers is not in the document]
ochaloup: Ok, I changed the code in this way.

pkg/controller/util/wildfly_mgmt.go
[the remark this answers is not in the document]
ochaloup: Ok. My concern is that the operator should be the most independent as possible. My vision would be possibility to run operator with s2i (that brings some more features) but still being possible to run it *without* s2i (that provides basic features - like scaling, ensures data consistency...). 

As you consider the s2i necessary (just I'm not sure what are those features where s2i is the requirements in current operator) I'm going to change the code to run only the *ssh remote calls* and won't be using the http management calls.

pkg/controller/wildflyserver/wildflyserver_controller.go
ochaloup: Yes, the EJB communication goes over the HTTP upgrade but there is possible to run EJB communication only over HTTP (https://github.com/wildfly/wildfly-http-client).

I'm not sure how many users used that option but at least no one with the transaction context propagation as it was not working (https://issues.jboss.org/browse/WFLY-12184)

ochaloup:
@jmesnil I updated the code to address (hopefully :) your points. Can you please take another look, please?

deploy/role.yaml
```diff
@@ -31,6 +32,12 @@ rules:
   - statefulsets
   verbs:
   - '*'
+- apiGroups:
```
jmesnil: I think `pods/log` should be put in the `apiGroups` just above (with `pods` and `pods/exec`)

pkg/controller/util/remote_ops.go
```diff
@@ -0,0 +1,238 @@
...
+	message, err := bufio.NewReader(conn).ReadString('\n')
+	if err != nil {
+		return "", fmt.Errorf("Error to get response for command %s sending to %s:%v, error: %v",
+			command, hostname, port, err)
+	}
+	return message, nil
+}
+
+// ReadPodLog reads log from the specified pod. Using Kubernetes REST Client to query the API
+func ReadPodLog(pod *corev1.Pod, logOptions *corev1.PodLogOptions) (io.ReadCloser, error) {
```
jmesnil: I think it should be `readPodLog`

pkg/controller/wildflyserver/wildflyserver_controller.go
```diff
@@ -504,20 +550,339 @@ func (r *ReconcileWildFlyServer) routeForWildFly(w *wildflyv1alpha1.WildFlyServe
...
+	scaleDownPodRecoveryPort, err := wildflyutil.GetTransactionRecoveryPort(scaleDownPod, jbossHome)
+	if err != nil || scaleDownPodRecoveryPort == 0 {
+		scaleDownPodRecoveryPort = defaultRecoveryPort
+	}
+	if err != nil {
+		reqLogger.Error(err, "Error on reading transaction recovery port with management command")
+	}
+	// With enabled recovery listener and the port, let's start the recovery scan
+	reqLogger.Info("Executing recovery scan", "Pod name", scaleDownPodName, "Pod IP", scaleDownPodIP, "Recovery port", scaleDownPodRecoveryPort)
+	_, err = wildflyutil.SocketConnect(scaleDownPodIP, scaleDownPodRecoveryPort, txnRecoveryScanCommand)
```
jmesnil: out of curiosity, why is the recovery scan not exposed as a management operation in the transaction subsystem?

pkg/controller/wildflyserver/wildflyserver_controller.go
```diff
@@ -504,20 +550,339 @@ func (r *ReconcileWildFlyServer) routeForWildFly(w *wildflyv1alpha1.WildFlyServe
...
+func getWildflyServerPodStatusByName(w *wildflyv1alpha1.WildFlyServer, podName string) *wildflyv1alpha1.PodStatus {
+	for _, podStatus := range w.Status.Pods {
+		if podName == podStatus.Name {
+			return &podStatus
+		}
+	}
+	return nil
+}
+
+func containsInMap(m map[string]string, v string) bool {
```
jmesnil: maybe we should move all these helper methods in a separate go file

jmesnil: maybe add a comment that you're checking if a map contains the `v` string in its *values*

pkg/controller/wildflyserver/wildflyserver_controller.go
```diff
@@ -33,15 +37,21 @@ import (
...
+	httpApplicationPort           int32 = 8080
+	httpManagementPort            int32 = 9990
+	defaultRecoveryPort           int32 = 4712
+	wildflyServerFinalizer              = "finalizer.wildfly.org"
+	wildflyHome                         = "/wildfly"
+	standaloneServerDataDirSuffix       = "/standalone/data"                // data directory where runtime data is saved
+	wftcDataDirName                     = "ejb-xa-recovery"                 // data directory where WFTC stores transaction runtime data
+	labelMarkerOperatedByService        = "wildfly.org/operated-by-service" // label used to remove a pod from receiving load from service during transaction recovery
+	txnRecoveryScanCommand              = "SCAN"                            // Narayana socket command to force recovery
+	reloadRetryCount                    = 10                                // number of retries when waiting for container reload is done (a retry is 1 second)
```
jmesnil: where are you spacing out every reload check for 1 second?

pkg/controller/wildflyserver/wildflyserver_controller.go
```diff
@@ -504,20 +550,339 @@ func (r *ReconcileWildFlyServer) routeForWildFly(w *wildflyv1alpha1.WildFlyServe
...
+			"Management command: %v, JSON response: %v", scaleDownPodName, wildflyutil.MgmtOpTxnEnableRecoveryListener, jsonResult)
+	}
+	// Enabling the recovery listner may require the server being reloaded
+	isReloadRequired := wildflyutil.ReadJSONDataByIndex(jsonResult["response-headers"], "operation-requires-reload")
+	if isReloadRequiredBool, _ := strconv.ParseBool(isReloadRequired); isReloadRequiredBool {
+		jsonResult, err = wildflyutil.ExecuteMgmtOp(scaleDownPod, jbossHome, wildflyutil.MgmtOpReload)
+		if err != nil {
+			return false, "", fmt.Errorf("Cannot reload application container for scaling down pod %v, error: %v", scaleDownPodName, err)
+		}
+		for serverStateCheckCounter := 0; err != nil && serverStateCheckCounter < reloadRetryCount; serverStateCheckCounter++ {
```
jmesnil: Does that mean that the server has 10 seconds to reload?
 My comment at https://github.com/wildfly/wildfly-operator/pull/54#discussion_r311515946 still stands (or maybe you could use a goroutine at least)

pkg/controller/wildflyserver/wildflyserver_controller.go
ochaloup: As this was not my design decision I'm not 100% sure about that. But in general I can say that forcing the recovery is not a standard operation that is about to be offered to a usual user. It's internal Narayana procedure which is not expected to be touched. It's helpful only for a person who really understands how Narayana works and what happens when it's called. It's useful for testing and for integration purposes. But I can't imagine that would be useful for an ordinary developer to force the recovery processing.
What a developer/admin/an op person can change is the period for the processing. This is currently possible only via system properties (not available in model).

As I wonder what are all the reasons behind  I created a forum question about this at: https://developer.jboss.org/message/990677#990677

pkg/controller/wildflyserver/wildflyserver_controller.go
ochaloup: I moved only the generic methods. Maybe some more could be considered. But as I'm looking at them now they seem to me belonging to the controller class based on their functionality.

pkg/controller/wildflyserver/wildflyserver_controller.go
ochaloup: thanks. this is a wrong comment. it was unintentionally left here from a previous version. I'm going to fix the comment.

pkg/controller/wildflyserver/wildflyserver_controller.go
ochaloup: The server has 10 seconds to reload. After that time it's considered to fail on recovery  processing. The reconcile loop is retried and the server is either already ready to response or it gets another 10 seconds to be reloaded. Maybe this 10 second is not necessary and the reconcile loop just takes care.

pkg/controller/wildflyserver/wildflyserver_controller.go
ochaloup: as it's not relevant for headless service where clustering responses is managed by application itself (in ejb case by the app server)

deploy/crds/wildfly_v1alpha1_wildflyserver_crd.yaml
```diff
@@ -87,11 +87,24 @@ spec:
...
+                      during scale down the expected values are represented by the
+                      PodState* constants  Read-only.
+                    type: string
                 required:
                 - name
                 - podIP
+                - state
                 type: object
               type: array
+            scalingdownPods:
```
simkam: please update also documentation in doc dir?

ochaloup: I added text there. Thanks for the hint.

ochaloup:
Updated with several tweaks and addressing https://issues.jboss.org/browse/WFWIP-183

doc/apis.adoc
```diff
@@ -48,6 +48,7 @@ It uses a `StatefulSet` with a pod spec that mounts the volume specified by `sto
 | `env` | List of environment variable present in the containers | []https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#envvar-v1-core[corev1.EnvVar] | false
 | `disableHTTPRoute`| Disable the creation a route to the HTTP port of the application service | bool | false
 | `sessionAffinity`| If connections from the same client ip are passed to the same WildFlyServer instance/pod each time | bool | false
+| `jbossHome` | Definition of the JBOSS_HOME directory in the container. It's used by the scale down processing which runs jboss-cli scripts. | string | false
```
pkremens: This one seems to be a same user story as https://github.com/wildfly/wildfly-operator/issues/51#issuecomment-513712928 I don't think we should expose this to end user

ochaloup: ok,, I see. Then @jmesnil defines the strategy on reviewing this PR. My perspective is it's good thing to give the user to customize the setup. Different images can be used to get WFLY running. As the S2I is the base image it does not necessarily mean not giving chance to run other images as well.

jmesnil: -1 the location of the app server is an implementation detail that should not be exposed to the user.

It differs between WildFly and EAP so the operation must cope with this (and we'll use internal env var to configure it) but this must not leak to the user API (the `WildFlyServer` CRD)

@ochaloup using S2I *is* a requirement. if users creates their images from another way (using vanilla WildFly image), it might or might not work but this is not a concern. To use WildFly (resp. EAP) Operator, the user *must* built their image using WildFly (resp. EAP) S2I image.

pkg/apis/wildfly/v1alpha1/wildflyserver_types.go
```diff
@@ -12,17 +12,19 @@ import (
...
+	SessionAffinity     bool                     `json:"sessionAffinity,omitempty"` // TODO: how to use this?
 	DisableHTTPRoute    bool                     `json:"disableHTTPRoute,omitempty"`
 	StandaloneConfigMap *StandaloneConfigMapSpec `json:"standaloneConfigMap,omitempty"`
 	Storage             *StorageSpec             `json:"storage,omitempty"`
 	ServiceAccountName  string                   `json:"serviceAccountName,omitempty"`
 	EnvFrom             []corev1.EnvFromSource   `json:"envFrom,omitempty"`
 	// Env contains environment variables for the containers running the WildFlyServer application
 	Env []corev1.EnvVar `json:"env,omitempty"`
+	// JBossHome defines place where jboss distribution is located at the docker image defined by ApplicationImage
+	JbossHome string `json:"jbossHome,omitempty"`
```
jmesnil: -1 (as explained above)

pkg/controller/util/go_utilities.go
```diff
@@ -0,0 +1,36 @@
+package util
+
+// ContainsInMap returns true if the map m contains at least on of the string
```
jmesnil: typo: `at least one`

test/framework/common.go
```diff
@@ -17,30 +17,37 @@ import (
...
+		t.Fatal(err)
+	}
+}
+
+func wildflyTestSetup(t *testing.T) (*framework.TestCtx, *framework.Framework) {
+	ctx := framework.NewTestCtx(t)
 	err := ctx.InitializeClusterResources(&framework.CleanupOptions{TestContext: ctx, Timeout: cleanupTimeout, RetryInterval: cleanupRetryInterval})
 	if err != nil {
-		t.Fatalf("failed to initialize cluster resources: %v", err)
+		defer ctx.Cleanup()
```
jmesnil: that does not look correct. If you are deferring the cleanup, it will be cleaned up when the method returns.

test/framework/common.go
```diff
@@ -162,3 +153,54 @@ func wildflyClusterViewTest(t *testing.T, f *framework.Framework, ctx *framework
 
 	return WaitUntilClusterIsFormed(f, t, wildflyServer, "clusterbench-0", "clusterbench-1")
 }
+
+// WildFlySmokeRecoveryScaledownTest runs recovery scale down operation
+func WildFlySmokeRecoveryScaledownTest(t *testing.T, applicationTag string) {
```
jmesnil: is the test actually performing recovery during scale down or is it just testing scaling down?

pkg/controller/wildflyserver/wildflyserver_controller.go
```diff
@@ -249,64 +327,51 @@ func (r *ReconcileWildFlyServer) Reconcile(request reconcile.Request) (reconcile
...
-}
-
-// matches checks if the envVar from the WildFlyServerSpec matches the same env var from the container.
-// If it does not match, it updates the container EnvVar with the fields from the WildFlyServerSpec and return false.
-func matches(container *v1.Container, envVar corev1.EnvVar) bool {
-	for index, e := range container.Env {
-		if envVar.Name == e.Name {
-			if !reflect.DeepEqual(envVar, e) {
-				container.Env[index] = envVar
+	// sorting pods by number in the name
```
jmesnil: this one could go in a utility method

pkg/controller/wildflyserver/wildflyserver_controller.go
```diff
@@ -249,64 +327,51 @@ func (r *ReconcileWildFlyServer) Reconcile(request reconcile.Request) (reconcile
...
 }
 
 // statefulSetForWildFly returns a wildfly StatefulSet object
 func (r *ReconcileWildFlyServer) statefulSetForWildFly(w *wildflyv1alpha1.WildFlyServer) *appsv1.StatefulSet {
-	ls := labelsForWildFly(w)
 	replicas := w.Spec.Size
 	applicationImage := w.Spec.ApplicationImage
 	volumeName := w.Name + "-volume"
+	labesForActiveWildflyPod := labelsForWildFly(w)
+	labesForActiveWildflyPod[markerOperatedByService] = "active"
```
jmesnil: `"active"` is referenced several time, maybe you could put it in a const

test/e2e/16.0/wildflyserver_test.go
```diff
@@ -43,3 +36,7 @@ func wildFlyBasicTest(t *testing.T) {
 func wildFlyClusterTest(t *testing.T) {
 	wildflyframework.WildFlyClusterTest(t, "16.0")
 }
+
+func wildFlySmokeRecoveryScaledownTest(t *testing.T) {
```
jmesnil: you can remove this test (I kept the WildFly 16 tests just in case) but the WildFly Operator really started working with WildFLy 17 (and its new S2I image)

doc/apis.adoc
ochaloup: I see and understand. Just still I think making the life easier as well for the non-S2I would be nice. Either way I deleted the setup property.

test/framework/common.go
ochaloup: This should be correct as the `defer.Cleanup()` is called just before the `t.Fatalf` is invoked. Saying differently it will be called when the method returns but the method returns just at the next call with error for whole test.

test/framework/common.go
ochaloup: It's only testing scale down (which means there is run recovery processing in the background). Do you consider changing name to like `WildflyScaleDownTest` as a better idea?

ochaloup:
@jmesnil updated based on your comments

ochaloup:
The latest changes are about to sync with the `upstream/master` (rebase was done).

pkg/controller/wildflyserver/wildflyserver_controller.go
```diff
@@ -31,18 +36,27 @@ import (
...
 
 const (
-	httpApplicationPort int32 = 8080
-	httpManagementPort  int32 = 9990
+	controllerName                      = "wildflyserver-controller"
+	httpApplicationPort           int32 = 8080
+	httpManagementPort            int32 = 9990
+	defaultRecoveryPort           int32 = 4712
+	wildflyServerFinalizer              = "finalizer.wildfly.org"
+	wildflyHome                         = "/opt/jboss/wildfly"
```
jmesnil: Please revert this change, in #55, I fixed the server location so that ` jbossHome` points to the correct server location and `standaloneServerDataDirPath` is correct.

pkg/controller/wildflyserver/wildflyserver_controller.go
```diff
@@ -31,18 +36,27 @@ import (
...
+	standaloneServerDataDir             = "/standalone/data"                // data directory where runtime data is saved
```
jmesnil: can be removed (see above)

pkg/controller/wildflyserver/wildflyserver_controller.go
```diff
@@ -347,7 +393,7 @@ func (r *ReconcileWildFlyServer) statefulSetForWildFly(w *wildflyv1alpha1.WildFl
 						ReadinessProbe: createReadinessProbe(),
 						VolumeMounts: []corev1.VolumeMount{{
 							Name:      volumeName,
-							MountPath: standaloneServerDataDirPath,
+							MountPath: r.getJbossHome() + standaloneServerDataDir,
```
jmesnil: please revert

pkg/controller/wildflyserver/wildflyserver_controller.go
```diff
@@ -435,7 +481,7 @@ func (r *ReconcileWildFlyServer) statefulSetForWildFly(w *wildflyv1alpha1.WildFl
 		})
 		statefulSet.Spec.Template.Spec.Containers[0].VolumeMounts = append(statefulSet.Spec.Template.Spec.Containers[0].VolumeMounts, corev1.VolumeMount{
 			Name:      "standalone-config-volume",
-			MountPath: jbossHome + "/standalone/configuration/standalone.xml",
+			MountPath: r.getJbossHome() + "/standalone/configuration/standalone.xml",
```
jmesnil: please revert

pkg/controller/wildflyserver/wildflyserver_controller.go
```diff
@@ -505,20 +548,395 @@ func (r *ReconcileWildFlyServer) routeForWildFly(w *wildflyv1alpha1.WildFlyServe
...
+		return false, nil
+	}
+	return true, nil
+}
+
+func (r *ReconcileWildFlyServer) checkRecovery(reqLogger logr.Logger, scaleDownPod *corev1.Pod, w *wildflyv1alpha1.WildFlyServer) (bool, string, error) {
+	scaleDownPodName := scaleDownPod.ObjectMeta.Name
+	scaleDownPodIP := scaleDownPod.Status.PodIP
+	scaleDownPodRecoveryPort := defaultRecoveryPort
+	jbossHome := r.getJbossHome()
```
jmesnil: please delete (`jbossHome` is now set properly)

pkg/controller/wildflyserver/wildflyserver_controller.go
```diff
@@ -505,20 +548,395 @@ func (r *ReconcileWildFlyServer) routeForWildFly(w *wildflyv1alpha1.WildFlyServe
...
+		retString := fmt.Sprintf("WildFly Transaction Client data dir is not empty and scaling down of the pod '%v' will be retried."+
+			"Wildfly Transacton Client data dir path '%v', output listing: %v",
+			scaleDownPodName, jbossHome+standaloneServerDataDir+"/"+wftcDataDirName, commandResult)
+		return false, retString, nil
+	}
+	return true, "", nil
+}
+
+func (r *ReconcileWildFlyServer) setupRecoveryPropertiesAndRestart(reqLogger logr.Logger, scaleDownPod *corev1.Pod, w *wildflyv1alpha1.WildFlyServer) (bool, error) {
+	jbossHome := r.getJbossHome()
```
jmesnil: please delete

pkg/controller/wildflyserver/wildflyserver_controller.go
```diff
@@ -505,20 +548,395 @@ func (r *ReconcileWildFlyServer) routeForWildFly(w *wildflyv1alpha1.WildFlyServe
...
+			return false, fmt.Errorf("Failed to update pod annotations, pod name %v, annotations to be set %v, error: %v",
+				scaleDownPodName, scaleDownPod.Annotations, err)
+		}
+	} else {
+		reqLogger.Info("Recovery properties for app server at pod was already defined. Skipping server restart.", "Pod Name", scaleDownPodName)
+	}
+	return true, nil
+}
+
+func (r *ReconcileWildFlyServer) getJbossHome() string {
```
jmesnil: please delete

ochaloup:
Small tweak to fix an issue occurred during investigation of JBEAP-17347.
Still ready for review.

pkg/controller/wildflyserver/wildflyserver_controller_test.go
```diff
@@ -101,7 +102,7 @@ func TestEnvUpdate(t *testing.T) {
 		},
 		Spec: wildflyv1alpha1.WildFlyServerSpec{
 			ApplicationImage: applicationImage,
-			Size:             replicas,
+			Size:             0,
```
jmesnil: why do you create the app without any pods?

pkg/controller/wildflyserver/wildflyserver_controller.go
```diff
@@ -196,63 +306,123 @@ func (r *ReconcileWildFlyServer) Reconcile(request reconcile.Request) (reconcile
...
-			reqLogger.Error(err, "Failed to update pods in WildFlyServer status.")
+			reqLogger.Error(err, "Failed to update WildFlyServer status.")
 			return reconcile.Result{}, err
 		}
+		requeue = true
 	}
-	if requeue {
-		return reconcile.Result{Requeue: true}, nil
+
+	mustReconcileStatefulset, requeueStatefulset, err := r.checkStatefulSet(wildflyServer, foundStatefulSet, podList)
```
jmesnil: I'm not sure if that is efficient to check update of the StatefulSet after creating all the resources (service, route) that may depend on it.

My idea was to keep the reconcile loop as linear as possible

```
1. check if the statefulset exists
2. if it exists, does it need to be updated?
  2.a yes, update it and requeue
  2.b no, continue
// at this point, we know that the statefulset spec is up to date with the wildflyserver
// any resources will be created from an up to state statefuls
...
```

pkg/controller/wildflyserver/wildflyserver_controller_test.go
ochaloup: The test does not need any pod for check it works. The reason why it does not work is that the operator waits for the all pods to up and running before it permits other operations to be processed.
The reconcile loop is defined here:
https://github.com/wildfly/wildfly-operator/pull/54/files#diff-d96e34894c3a7c83e1eb1ad11a0ab147R224
The handling is here because of the recovery processing lists all the available pods and then  works above it.

pkg/controller/wildflyserver/wildflyserver_controller.go
ochaloup: I see, fair point. I will change this.

ochaloup:
@jmesnil I updated the PR. The state of StatefulSet is changed as it was before (just after updating the spec).

PR #67, opened 2019-09-11
[#66] Changes to WildFlyServer Spec are not reflected in StatefulSet

### LINKED ISSUE #66: Changes to WildFlyServer Spec are not reflected in StatefulSet
1. create WildFlyServer without PVC

```
kind: WildFlyServer
  name: operator-added
spec:
  ...
```

2. In runtime add storage into WildFly CR definition

```
$ oc edit pvc

kind: WildFlyServer
metadata:
  name: operator-added
spec:
  ...
  storage:
    volumeClaimTemplate:
      apiVersion: v1
      kind: PersistentVolumeClaim
      metadata:
        annotations: {}
        labels: {}
        name: operator-added-volume
      spec:
        accessModes:
          - ReadWriteMany
        resources:
          limits: {}
          requests:
            storage: 1Mi
```

3. There is no new pvc object in kubernetes 

`$ oc get pvc`

jmesnil:
Add `wildfly.org/wildfly-server-generation` annotation to the
statefulset to track the generation of the `WildFlyServer` that created
it.

If the generation does not match, update or delete the `statefulset` so
that its spec will be update according to the latest generation of the
`WildFlyServer`.

This fixes #66

### DIFF
--- a/pkg/controller/wildflyserver/wildflyserver_controller.go
+++ b/pkg/controller/wildflyserver/wildflyserver_controller.go
@@ -148,16 +148,11 @@ func (r *ReconcileWildFlyServer) Reconcile(request reconcile.Request) (reconcile
 		return reconcile.Result{}, err
 	}
 
-	// check if the stateful set is up to date with the WildFlyServerSpec
-	if checkUpdate(wildflyServer, foundStatefulSet) {
-		err = r.client.Update(context.TODO(), foundStatefulSet)
-		if err != nil {
-			reqLogger.Error(err, "Failed to update StatefulSet.", "StatefulSet.Namespace", foundStatefulSet.Namespace, "StatefulSet.Name", foundStatefulSet.Name)
-			return reconcile.Result{}, err
-		}
-
-		// Spec updated - return and requeue
-		return reconcile.Result{Requeue: true}, nil
+	mustReconcile, requeue, err := r.checkStatefulSet(wildflyServer, foundStatefulSet)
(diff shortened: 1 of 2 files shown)


pkg/controller/wildflyserver/wildflyserver_controller.go
```diff
@@ -253,41 +248,66 @@ func (r *ReconcileWildFlyServer) Reconcile(request reconcile.Request) (reconcile
...
+					if err = r.client.Delete(context.TODO(), foundStatefulSet); err != nil {
+						log.Error(err, "Failed to Delete StatefulSet.", "StatefulSet.Namespace", foundStatefulSet.Namespace, "StatefulSet.Name", foundStatefulSet.Name)
+						return true, false, err
+					}
+					log.Info("Deleting StatefulSet that is not up to date with the WildFlyServer StorageSpec", "StatefulSet.Namespace", foundStatefulSet.Namespace, "StatefulSet.Name", foundStatefulSet.Name)
+					return true, true, nil
+				}
+
+				// all other changes are in the spec Template or Replicas and the statefulset can be updated
+				foundStatefulSet.Spec.Template = statefulSet.Spec.Template
```
ochaloup: @jmesnil : just checking as I don't understand here if the changes will be reflected in pods

As the template definition is base for the pod creation, does the changing of template cause changing of the currently active pods? Does not need the spec template change removing the statefulset and recreating it?

Issue #72, opened 2019-09-17
Transaction recovery handling finalizer does not permit wildflyserver to be terminated when there is no active pod


ochaloup:
Finalizer makes the operator removal being stuck when there is no active pod under operator management. The operator repeats error message[1] and never shuts down.
The activity may end up with panic.

```
{"level":"info","ts":1568704825.2354705,"logger":"controller_wildflyserver","msg":"WildflyServer is marked for deletion. Waiting for finalizers to clean the workspace","Request.Namespace":"default","Request.Name":"quickstart"}
{"level":"info","ts":1568704825.2425647,"logger":"controller_wildflyserver","msg":"Transaction recovery scaledown processing","Request.Namespace":"default","Request.Name":"quickstart","Pod Name":"quickstart-0","IP Address":"172.17.0.6"}
{"level":"info","ts":1568704825.2728443,"logger":"controller_wildflyserver","msg":"Enabling recovery listener for processing scaledown at quickstart-0","Request.Namespace":"default","Request.Name":"quickstart"}
[...]

PR #74, opened 2019-09-19
Refactor resources

jmesnil:
* Move underlying resources created by the operator in their own packages
(statefulset, services, routes)
* Add actions.go to group CRUD actions on the resources
* Add server generation label to resources that are created/updated


jmesnil:
@ochaloup fiy, I'm refactoring the code to split resource CRUD operations and creation from the controller file.
The controller file should only deal with the reconcile loop and other tasks be splitter in smaller dedicated packages.

I'll wait for your fix on #72 before I merge to avoid conflict in your PR

ochaloup:
@jmesnil thanks Jeff for patience. Unfortunately I've found a different small trouble on scaledown and I have an issue to fix it correctly. I need to leave now and I will be ready with my PR tomorrow in the morning.

PR #75, opened 2019-09-20
[issue72] pending container scaledown

### LINKED ISSUE #72: Transaction recovery handling finalizer does not permit wildflyserver to be terminated when there is no active pod

ochaloup:
fixes #72 

Fixing the issues found for scaledown processing.
The scaledown could hang in some cases, a panic exception could be got when pod annotations were not ready to be used and a wrong error was reported to the user during processing the finalizer.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -42,6 +42,7 @@ require (
 	github.com/sirupsen/logrus v1.4.1 // indirect
 	github.com/spf13/pflag v1.0.3
 	github.com/stretchr/testify v1.3.0
+	github.com/tevino/abool v0.0.0-20170917061928-9b9efcf221b5
 	go.uber.org/atomic v1.3.2 // indirect
 	go.uber.org/multierr v1.1.0 // indirect
 	go.uber.org/zap v1.9.1 // indirect
--- a/go.sum
+++ b/go.sum
@@ -270,6 +270,8 @@ github.com/stretchr/testify v1.2.2/go.mod h1:a8OnRcib4nhh0OaRAV+Yts87kKdq0PP7pXf
 github.com/stretchr/testify v1.3.0 h1:TivCn/peBQ7UY8ooIcPgZFpTNSz0Q2U6UrFlUfqbe0Q=
 github.com/stretchr/testify v1.3.0/go.mod h1:M5WIy9Dh21IEIfnGCwXGc5bZfKNJtfHm1UVUgZn+9EI=
 github.com/tarm/serial v0.0.0-20180830185346-98f6abe2eb07/go.mod h1:kDXzergiv9cbyO7IOYJZWg1U88JhDg3PB6klq9Hg2pA=
+github.com/tevino/abool v0.0.0-20170917061928-9b9efcf221b5 h1:hNna6Fi0eP1f2sMBe/rJicDmaHmoXGe1Ta84FPYHLuE=
+github.com/tevino/abool v0.0.0-20170917061928-9b9efcf221b5/go.mod h1:f1SCnEOt6sc3fOJfPQDRDzHOtSXuTtnz0ImG9kPRDV0=
 go.opencensus.io v0.18.0/go.mod h1:vKdFvxhtzZ9onBp9VKHK8z/sRpBMnKAsufL7wlDrCOA=
 go.opencensus.io v0.20.1/go.mod h1:6WKK9ahsWS3RSO+PY9ZHZUfv2irvY6gN279GOPZjmmk=
 go.opencensus.io v0.20.2 h1:NAfh7zF0/3/HqtMvJNZ/RFrSlCE6ZTlHmKfhL/Dm1Jk=
--- a/pkg/controller/util/go_utilities.go
+++ b/pkg/controller/util/go_utilities.go
@@ -72,3 +72,18 @@ func SortPodListByName(podList *corev1.PodList) *corev1.PodList {
 	})
 	return podList
 }
+
+// MapMerge merges the two maps together and returns the result.
+// If one of them is nil then is ommitted from the merged result
+// If both maps are null then an empty initialized map is returned
+// The second map rewrites data of the first map in the result if there are such
+func MapMerge(firstMap map[string]string, secondOverwritingMap map[string]string) map[string]string {
+	returnedMap := make(map[string]string)
+	for v, k := range firstMap {
+		returnedMap[v] = k
+	}
+	for v, k := range secondOverwritingMap {
+		returnedMap[v] = k
+	}
+	return returnedMap
+}
--- a/pkg/controller/util/remote_ops.go
+++ b/pkg/controller/util/remote_ops.go
@@ -27,6 +27,8 @@ var (
 	timestampFromLogLine = regexp.MustCompile(`^[0-9]{4}-[0-9]{2}-[0-9]{2}[^ ]+`)
 	// start of the time timestamp
 	startOfTheTimeTimestamp = time.Unix(0, 0)
+	// socket dial timeout
+	socketDialTimeout = 10 * time.Second
 )
 
 //ExecRemote executes a command inside the remote pod
@@ -87,7 +89,7 @@ func ExecRemote(pod *corev1.Pod, command string) (string, error) {
 func SocketConnect(hostname string, port int32, command string) (string, error) {
 	// connect to socket
 	toConnectTo := fmt.Sprintf("%v:%v", hostname, port)
-	conn, err := net.Dial("tcp", toConnectTo)
+	conn, err := net.DialTimeout("tcp", toConnectTo, socketDialTimeout)
 	if err != nil {
 		return "", fmt.Errorf("Cannot process TCP connection to %v, error: %v",
 			toConnectTo, err)
--- a/pkg/controller/util/wildfly_mgmt.go
+++ b/pkg/controller/util/wildfly_mgmt.go
@@ -22,8 +22,8 @@ var (
 	MgmtOpReload = ":reload()"
 	// MgmtOpRestart is a JBoss CLI command for restarting WFLY server
 	MgmtOpRestart = ":shutdown(restart=true)"
-	// MgmtOpTxnEnableRecoveryListener is a JBoss CLI command for enabling txn recovery listener
-	MgmtOpTxnEnableRecoveryListener = "/subsystem=transactions:write-attribute(name=recovery-listener, value=true)"
+	// MgmtOpTxnCheckRecoveryListener is a JBoss CLI command for enabling txn recovery listener
+	MgmtOpTxnCheckRecoveryListener = "/subsystem=transactions:read-attribute(name=recovery-listener)"
 	// MgmtOpTxnProbe is a JBoss CLI command for probing transaction log store
 	MgmtOpTxnProbe = "/subsystem=transactions/log-store=log-store:probe()"
 	// MgmtOpTxnRead is a JBoss CLI command for reading transaction log store
--- a/pkg/controller/util/wildfly_mgmt.go
+++ b/pkg/controller/util/wildfly_mgmt.go
@@ -202,7 +202,7 @@ func ExecuteOpAndWaitForServerBeingReady(mgmtOp string, pod *corev1.Pod, jbossHo
 		return false, fmt.Errorf("Cannot run operation '%v' at application container for down pod %s, error: %v", mgmtOp, podName, err)
 	}
 	if !IsMgmtOutcomeSuccesful(jsonResult) {
-		return false, fmt.Errorf("Not succefully runnin management operation '%v' for pod %s. JSON output: %v",
+		return false, fmt.Errorf("Not succefully running management operation '%v' for pod %s. JSON output: %v",
 			mgmtOp, podName, jsonResult)
 	}
 	for serverStateCheckCounter := 0; err != nil && serverStateCheckCounter < reloadRetryCount; serverStateCheckCounter++ {
--- a/pkg/controller/wildflyserver/wildflyserver_controller.go
+++ b/pkg/controller/wildflyserver/wildflyserver_controller.go
@@ -11,6 +11,7 @@ import (
 	"sync"
 
 	"github.com/go-logr/logr"
+	"github.com/tevino/abool"
 	wildflyv1alpha1 "github.com/wildfly/wildfly-operator/pkg/apis/wildfly/v1alpha1"
 	wildflyutil "github.com/wildfly/wildfly-operator/pkg/controller/util"
 
@@ -180,13 +181,17 @@ func (r *ReconcileWildFlyServer) Reconcile(request reconcile.Request) (reconcile
 			//   this will be retried at next reconciliation.
 			r.recorder.Event(wildflyServer, corev1.EventTypeNormal, "WildFlyServerDeletion",
 				"Scaledown processing started before removal. Consult operator log for status.")
-			if err := r.finalizeWildflyServer(reqLogger, wildflyServer); err != nil {
+			requeue, err := r.finalizeWildflyServer(reqLogger, wildflyServer)
+			if err != nil {
 				return reconcile.Result{}, err
+			}
+			if requeue {
+				return reconcile.Result{Requeue: true}, nil
 			}
 
 			// Remove WildflyServer. Once all finalizers have been removed, the object will be deleted.
 			wildflyServer.SetFinalizers(wildflyutil.RemoveFromList(wildflyServer.GetFinalizers(), wildflyServerFinalizer))
-			err := r.client.Update(context.TODO(), wildflyServer)
+			err = r.client.Update(context.TODO(), wildflyServer)
 			if err != nil {
 				return reconcile.Result{}, err
 			}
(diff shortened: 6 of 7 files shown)


jmesnil:
@ochaloup Your fix is not working: https://gist.github.com/jmesnil/53b90ca66db0100151587c746c81dead

`kubectl delete` is still hanging when the `WildFlyServer` is deleted.

To reproduce:
* I pulled your branch
* modify deploy/operator.yaml to point to my repo on quay.io
* `$ make run-minikube`
* `$ kubectl apply -f deploy/crds/quickstart-cr.yaml`
* wait for both pods to be started
* `$ kubectl delete -f deploy/crds/quickstart-cr.yaml`
=> it is hanging

jmesnil:
@ochaloup I have a silly question but why isn't the recovery listener always enabled?

When we shutdown a pod, we may have to reload the server just to be able to enable the recovery listener. But what happens to the transactions when the server is reloaded?

ochaloup:
@jmesnil 

> I have a silly question but why isn't the recovery listener always enabled?

The recovery listener is not always enabled as it's not expected to be used by user. It's the still the same answer, unfortunately ;-) Up to this time these commands were not expected to be used during the app development. It was about more for testing. We do not mention these options at the product documentations etc.
The s2i configuration probably enables the recovery listener so it's highly probable that the reload won't be needed. But still we need to be defensive as we don't know what are the changes that the user does.

> When we shutdown a pod, we may have to reload the server just to be able to enable the recovery listener. But what happens to the transactions when the server is reloaded?

The active in-memory transactions are just stopped and will be rolled-back (either by fact they are shutdown or by timeout at the remote resource). That should be ok as the pod is going to be removed.
The prepared transactions - the ones which are persisted - will be processed by recovery.

Agree, it's not ideal and it could be done better if there will be an special CLI operation that would provide the unfinished transaction directly. See my comment at https://github.com/wildfly/wildfly-operator/issues/72#issuecomment-533467408

jmesnil:
> The s2i configuration probably enables the recovery listener so it's highly probable that the reload won't be needed. But still we need to be defensive as we don't know what are the changes that the user does.

The Operator requires S2I. So we can always make sure that the S2I image has the transaction recovery enabled.

ochaloup:
@jmesnil I don't agree. You never know if the user does some changes in his `standalone.xml`, either with jboss cli script or by uploading his own `stanalone.xml`. We need to be defensive here and check here if it's enabled.

jmesnil:
@ochaloup If S2I image is used, the configuration of standalone.xml is controlled (and from the operator, we can always set any env var to ensure that transaction recovery listener is enabled).

If users bring their own standalone.xml, sure they can do what they want and we have to test whether tx recovery is enabled. If that's not the case, we should just log an error (saying that listener must be enabled) and abort

ochaloup:
@jmesnil The code currently checks if the tx recovery is enabled. If it's enabled then it does nothing. If it's not enabled then it proceed with setting it up.
Is there something wrong in this process?

jmesnil:
Regarding the issue, the code hangs because the connection to the tx recovery listener can not be done.
If I use busybox to connect from inside the cluster, I got:

```
# telnet 172.17.0.6:4712
telnet: can't connect to remote host (172.17.0.6): Connection refused
```

If I do a telnet on 8080, I can get the HTTP response so the server is Alice and well

jmesnil:
> The code currently checks if the tx recovery is enabled. If it's enabled then it does nothing. If it's not enabled then it proceed with setting it up.

Note that this is not the place where this PR is hanging so I don't want to side track it but I think you're making your life harder.
Since we use S2I, we can make sure that WildFly is started with the correct configuration (and the tx recovery is enabled).

Reloading the server while we are trying to shut it down is unfortunate.
We could:
* check that tx recovery listener is enabled when the server is started. if it is, good. If it is not, we can warn the user or abort *before* the actual tx recovery kicks in when the pod is shutdown.
* check that tx recovery listener is enabled (and it will if users uses S2I as required), otherwise we abort
* regarding the hanging issue, we can preemptively open a connection to the tx recovery port and disable tx recovery if connection can not be opened

Reloading the server while shutting down the pod seems to open a Pandora box of issue as we lose the immutability offered by the container platform.

ochaloup:
@jmesnil I understand the point on immutability it make sense. I consider the right solution here the CLI command which would provide the information without need of container reconfiguration. I admit I should think about this this way before.
Until then I consider the reload the lesser evil. Or at least my thinking is this - if tx recovery is not setup then scaledown processing for trasaction can do nothing and thus we either configure it or we looses any ability to assure the txn consistency. I think it's good to have this responsibility on the operator rather than saying in the documentation to users who changes `standalone.xml` configuration to not forget to enable txn listener.

jmesnil:
@ochaloup It is not the job of the operator to "fix" any misconfiguration from the application image.

If the user uses S2I, he does not change manually the standalone.xml but configure env vars for that.
We can make sure that the transaction subsystem comes with the tx recovery listener enabled.

BTW, why stopping at fixing the recovery tx listener configuration? What if the transactions subsystem itself is missing? Should you add it too? That's not sustainable.

We require S2I images to ensure that the WildFly configuration is "correct" and the operator can set any env var to make sure we have the bits we expect (e.g. transactions).

Do you agree?

ochaloup:
@jmesnil My perspective is different. The operator should help the user to run his application on top of the app server. If the user is not deep dived into the configuration of transaction subsystem he easily overlook to change the property. That way it's good to give him a hand.

> BTW, why stopping at fixing the recovery tx listener configuration? What if the transactions subsystem itself is missing? Should you add it too? That's not sustainable.

If the transaction subsystem is missing then recovery has no mean to be processed. It's a good point that the operator should be capable to find out the transaction subsystem is not available and thus it should not proceed with the recovery checks.

In summary we have a different opinions on this. I'm willing to go in your way.

Then the question is - what should happen when the recovery listener is not up?
It should be left the pod to be terminated and only log that data consistency could be in harm but there is misconfiguration and it's up to user to process that on his own?

ochaloup:
I updated the PR for fixing a trouble for WFWIP-206 and then I added a fix for your issue. If the scan is not possible to be processed then the cycle of setting-up recovery port should be rechecked. I can't say this fixes all the troubles as there is some trouble to run the {{jboss-cli.sh}} command for some reason. In my case they are run finally but at first they just fail.

jmesnil:
It is no longer hanging with the latest commit.
It only takes 3:40 minutes to stop a 2-pod cluster...

Could you please remove any code that modifies the server configuration and reload it? The server configuration is what it is, it is not the job of the operator to modify it.
If tx recovery is not enabled, the operator must not attempt to enable it.

ochaloup:
@jmesnil ok, I will

> If tx recovery is not enabled, the operator must not attempt to enable it.

What should be done then? If the tx recovery is not enabled then the pod will be permitted to be terminated even if there are some unfinished transactions? Or the pod won't never be terminated before the user will change the configuration for the tx recovery will be enabled?

My point is that if user doesn't setup tx recovery listener then he loses all the recovery guarantees.

jmesnil:
> If the tx recovery is not enabled then the pod will be permitted to be terminated even if there are some unfinished transactions

That's correct. If server is not configured to recover transactions, the server will be terminated.

ochaloup:
@jmesnil the latest commit makes the `tx-recovery-listener` as a required setup for the recovery would be processed. Otherwise the txn recovery is not run.

I was trying to understand what's the trouble with the REST calls etc. - and at least on my computer  - setting up the bigger REST call timeouts did not help.
What help was to add some resources to minikube. I run with 2GB of RAM first. When switch to 4GB the ssh remote calls over the REST API was processed without errors.

If this setup is fine for you then I would prefer to change your quickstart to contain the `tx-recovery-listener`. Otherwise the e2e tests do not test any txn recovery processing.
Is the `quay.io/jmesnil/wildfly-operator-quickstart:17.0` taken from WFLY s2i? I mean when the wfly s2i is changed to contain the galleon setup with tx recovery then the image of quickstart could be updated?

jmesnil:
> If this setup is fine for you then I would prefer to change your quickstart to contain the `tx-recovery-listener`. Otherwise the e2e tests do not test any txn recovery processing.
> Is the `quay.io/jmesnil/wildfly-operator-quickstart:17.0` taken from WFLY s2i?

Yes, it's taken from wildfy:17.0 S2I so there is nothing I can change here as its recovery listener is not enabled in the image.

For the e2e tests, you would need to create a new quay.io/jmesnil/wildfly-operator-quickstart:18.0 image built from wildfly:18.0 S2I (or that you build locally from the latest wildfly-s2i as WildFly 18 is not released yet).
Then we can add a new category of e2e for 18 that would test recovery.

jmesnil:
With the latest commit, an application with tx recovery disabled shut down in 7 seconds.

I have an error in the logs that I have not seen previously though:

```
{"level":"info","ts":1569242221.7286794,"logger":"controller_wildflyserver","msg":"Finalizer finished succesfully","Request.Namespace":"default","Request.Name":"quickstart","WildflyServer Namespace":"default","WildflyServer Name":"quickstart"}
{"level":"error","ts":1569242221.7401657,"logger":"kubebuilder.controller","msg":"Reconciler error","controller":"wildflyserver-controller","request":"default/quickstart","error":"Operation cannot be fulfilled on wildflyservers.wildfly.org \"quickstart\": StorageError: invalid object, Code: 4, Key: /registry/wildfly.org/wildflyservers/default/quickstart, ResourceVersion: 0, AdditionalErrorMsg: Precondition failed: UID in precondition: 6452435b-ddfe-11e9-9155-960f458eda08, UID in object meta: ","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/Users/jmesnil/go/pkg/mod/github.com/go-logr/zapr@v0.1.1/zapr.go:128\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/Users/jmesnil/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.1.12/pkg/internal/controller/controller.go:217\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func1\n\t/Users/jmesnil/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.1.12/pkg/internal/controller/controller.go:158\nk8s.io/apimachinery/pkg/util/wait.JitterUntil.func1\n\t/Users/jmesnil/go/pkg/mod/k8s.io/apimachinery@v0.0.0-20190221213512-86fb29eff628/pkg/util/wait/wait.go:133\nk8s.io/apimachinery/pkg/util/wait.JitterUntil\n\t/Users/jmesnil/go/pkg/mod/k8s.io/apimachinery@v0.0.0-20190221213512-86fb29eff628/pkg/util/wait/wait.go:134\nk8s.io/apimachinery/pkg/util/wait.Until\n\t/Users/jmesnil/go/pkg/mod/k8s.io/apimachinery@v0.0.0-20190221213512-86fb29eff628/pkg/util/wait/wait.go:88"}
```

I'll merge the PR as the server no longer hangs, thanks

Issue #33, opened 2019-07-04
Support scale subresource


jmesnil:
The WildFly Operator does not support the `kubectl scale ` command to scale up/down the number of replicas:

```
$ kubectl  scale  --replicas=3 wildflyserver/quickstart
Error from server (NotFound): the server could not find the requested resource (get wildflyservers quickstart)
```

If the verbosity is increased, the  wildflyserver/quickstart resource is actually found but not its `/scale` HTTP API

```
I0704 17:48:30.716361   55190 round_trippers.go:386] curl -k -v -XGET  -H "Accept:application/json" -H "User-Agent: oc/v1.11.0+d4cacc0 (darwin/amd64) kubernetes/d4cacc0" -H "Authorization: Bearer CcQa6C0m8BG7A0UHsraB1_W9xsmaUNW91dsnTTsOUdk" 'https://192.168.64.32:8443/apis/wildfly.org/v1alpha1/namespaces/myproject/wildflyservers/quickstart'
I0704 17:48:30.723857   55190 round_trippers.go:405] GET https://192.168.64.32:8443/apis/wildfly.org/v1alpha1/namespaces/myproject/wildflyservers/quickstart 200 OK in 7 milliseconds
...
I0704 17:48:30.728012   55190 round_trippers.go:386] curl -k -v -XGET  -H "User-Agent: oc/v1.11.0+d4cacc0 (darwin/amd64) kubernetes/d4cacc0" -H "Accept: application/json, */*" -H "Authorization: Bearer CcQa6C0m8BG7A0UHsraB1_W9xsmaUNW91dsnTTsOUdk" 'https://192.168.64.32:8443/apis/wildfly.org/v1alpha1/namespaces/myproject/wildflyservers/quickstart/scale'
I0704 17:48:30.729232   55190 round_trippers.go:405] GET https://192.168.64.32:8443/apis/wildfly.org/v1alpha1/namespaces/myproject/wildflyservers/quckstart/scale 404 Not Found in 1 milliseconds
```

See https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#scale-subresource

PR #89, opened 2019-09-26
[#33] Support scale subresources 

jmesnil:


jmesnil:
@ochaloup I've got one bad news and one good news.

The bad news was that the e2e tests run by the CI were using the image from quay.io and not the code from the PR. That's why you had some weird issues during development: the tests were using the code from the PR but the operator was running code from the old wildfly-operator 0.3.0

I notice that when changing the CRD in an incompatible way in this PR...

the good news is that I have fixed that so the e2e tests will run using a local operator (no longer deployed from an image).
However, there are now failures for the clustering and scaledown tests that may have been introduced at any point since the 0.3.0 releases...

I'm looking at the clustering failures but wonder if we should merge this PR now to make sure all futures PR will actually test their proposed changes.
@ochaloup would you agree?

ochaloup:
@jmesnil agree, +1 for merging this.
I think all futures PR should test their proposed changes correctly. When you merge this I will align my changes from https://github.com/wildfly/wildfly-operator/pull/90

Then about the testing I have another question about the wfly18 testing. It's if we can test the wfly18 image. I can create the `e2e` directory but there is need to be a quickstart docker image built from wfly18 'in progress branch' and published at the quay.io. I discussed how to built the image with Jean-Francois and it should be easier than it's describe in README.md at https://github.com/jmesnil/wildfly-operator-quickstart like just one `s2i` command (`s2i build . wildfly/wildfly-centos7:dev quay.io/ochaloup/wildfly-operator-quickstart:18.0-dev -e GALLEON_PROVISION_LAYERS=cloud-server`). Do you think you can push it that the test really checks basic scaledown processing? The issue is that wfly17 does not set-up the `tx-recovery-listener` and thus it processes no recovery cli calls.

PR #90, opened 2019-09-27
Scaledown recovery fixes

ochaloup:
Fixes for scaledown to work. These fixes came from issues brought by re-factoring from last days.

Fixes (besides other things) https://issues.jboss.org/browse/WFWIP-215 and operator issues for https://issues.jboss.org/browse/WFWIP-206

### DIFF
--- a/deploy/crds/wildfly_v1alpha1_wildflyserver_crd.yaml
+++ b/deploy/crds/wildfly_v1alpha1_wildflyserver_crd.yaml
@@ -12,6 +12,8 @@ spec:
     shortNames:
     - wfly
   scope: ""
+  subresources:
+    status: {}
   version: v1alpha1
   versions:
   - name: v1alpha1
--- a/pkg/controller/util/go_utilities.go
+++ b/pkg/controller/util/go_utilities.go
@@ -76,6 +76,12 @@ func SortPodListByName(podList *corev1.PodList) *corev1.PodList {
 	return podList
 }
 
+// CopyMap duplicates map as it creates new map and puts there all the
+//   key value pairs that were included in the first map
+func CopyMap(originalMap map[string]string) map[string]string {
+	return MapMerge(originalMap, nil)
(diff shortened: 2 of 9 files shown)


ochaloup:
@jmesnil Hi. Would you be so kind and review this PR?

deploy/crds/wildfly_v1alpha1_wildflyserver_crd.yaml
```diff
@@ -12,6 +12,8 @@ spec:
     shortNames:
     - wfly
   scope: ""
+  subresources:
```
jmesnil: since #88 I fixed code generation to make sure we don't edit this file manually.

Was this added by `make codegen`?

ochaloup: @jmesnil yes, that was added by `make codegen` but I needed to remove several other changes that running `codegen` on my machine changes the paths (like instead of `./pkg/apis/wildfly/v1alpha1.StandaloneConfigMapSpec` it changes to `github.com/wildfly/wildfly-operator/pkg/apis/wildfly/v1alpha1.StandaloneConfigMapSpec`)

I have not much idea why my `codegen` generates different results than the one on your machine.

Does your run of `make codegen` generates what is in the `wildfly_v1alpha1_wildflyserver_crd.yaml`? Why the `subresources` was removed from what was generated before #88?

jmesnil: I'm not sure why to be honest...

`status` is now generated as a top level field: https://github.com/wildfly/wildfly-operator/pull/88/files#diff-aa625220cf655743036e4c98524baaf2R23

I'm using `operator-sdk version: v0.10.0, commit: ff80b17737a6a0aade663e4827e8af3ab5a21170` to generate the files. Which version are you using?

PR #99, opened 2019-10-04
enhance error output of properties setup with data from ssh execution

ochaloup:
During testing we just found I unintentionally started to hide details on error message when CLI execution fails.

### DIFF
--- a/pkg/controller/wildflyserver/transaction_recovery.go
+++ b/pkg/controller/wildflyserver/transaction_recovery.go
@@ -174,15 +174,15 @@ func (r *ReconcileWildFlyServer) setupRecoveryPropertiesAndRestart(reqLogger log
 	if scaleDownPod.Annotations[markerRecoveryPropertiesSetup] == "" {
 		reqLogger.Info("Setting up back-off period and orphan detection properties for scaledown transaction reocovery", "Pod Name", scaleDownPodName)
 		setPeriodOps := fmt.Sprintf(wildflyutil.MgmtOpSystemPropertyRecoveryBackoffPeriod, "1") + "," + fmt.Sprintf(wildflyutil.MgmtOpSystemPropertyOrphanSafetyInterval, "1")
-		jsonResult, _ := wildflyutil.ExecuteMgmtOp(scaleDownPod, resources.JBossHome, setPeriodOps)
+		jsonResult, errExecution := wildflyutil.ExecuteMgmtOp(scaleDownPod, resources.JBossHome, setPeriodOps)
 
 		// command may end-up with error with status 'rolled-back' which means duplication, thus do not check for error as error could be possitive outcome
-
 		isOperationRolledBack := wildflyutil.ReadJSONDataByIndex(jsonResult, "rolled-back")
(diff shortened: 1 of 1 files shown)


pkg/controller/wildflyserver/transaction_recovery.go
```diff
@@ -174,15 +174,15 @@ func (r *ReconcileWildFlyServer) setupRecoveryPropertiesAndRestart(reqLogger log
...
 		reqLogger.Info("Setting up back-off period and orphan detection properties for scaledown transaction reocovery", "Pod Name", scaleDownPodName)
 		setPeriodOps := fmt.Sprintf(wildflyutil.MgmtOpSystemPropertyRecoveryBackoffPeriod, "1") + "," + fmt.Sprintf(wildflyutil.MgmtOpSystemPropertyOrphanSafetyInterval, "1")
-		jsonResult, _ := wildflyutil.ExecuteMgmtOp(scaleDownPod, resources.JBossHome, setPeriodOps)
+		jsonResult, errExecution := wildflyutil.ExecuteMgmtOp(scaleDownPod, resources.JBossHome, setPeriodOps)
 
 		// command may end-up with error with status 'rolled-back' which means duplication, thus do not check for error as error could be possitive outcome
-
 		isOperationRolledBack := wildflyutil.ReadJSONDataByIndex(jsonResult, "rolled-back")
 		isOperationRolledBackAsString, _ := wildflyutil.ConvertToString(isOperationRolledBack)
 		if !wildflyutil.IsMgmtOutcomeSuccesful(jsonResult) && isOperationRolledBackAsString != "true" {
```
jmesnil: you should also return false whenever `errExecution != nil`, right?

pkg/controller/wildflyserver/transaction_recovery.go
```diff
@@ -174,15 +174,15 @@ func (r *ReconcileWildFlyServer) setupRecoveryPropertiesAndRestart(reqLogger log
...
-			return false, fmt.Errorf("Setting up the back-off period and orphan detection properties by env property "+
-				"with operation '%s' at pod %s was not succesful. JSON output: %v", setPeriodOps, scaleDownPodName, jsonResult)
+			return false, fmt.Errorf("Error on setting up the back-off and orphan detection periods. "+
```
jmesnil: you should also return false whenever `errExecution != nil`, right?

ochaloup: No, at least now. This is meant to be a "tuning" for not running more jboss-cli commands than necessary. The command tries to setup the system variables and if it succeed then `errExecution == nil`.
If it fails there are two possibilities.
* The jboss-cli.sh command returns non-zero code when command fails. The command fails if it does not succeed. That's for example the case when the resource is already present and the code tries to set the variable again.
* The golang ssh remote execution returns non-zero code when command fails because of network issues etc.

For the code here it's a correct state when the golang ssh command runs correctly but when jboss-cli.sh command reported error because of the resource duplication.

That's why the `errExecution != nil` is not everytime considered as "`false`".

PR #106, opened 2019-10-04
[#105] Update S2I images for WildFly 18

### LINKED ISSUE #105: Update S2I images for WildFly 18
* update quickstarts & examples to use WildFly 18.0
* add e2e test for WildFly 18.0
* remove e2e tests for WildFly 16.0

jmesnil:
* missing images for clustering and postgresql examples
* update clustering example to use standard XML configuration

This fixes #105.


build/Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM registry.access.redhat.com/ubi7/ubi-minimal:latest
 
 ENV OPERATOR=/usr/local/bin/wildfly-operator \
-    JBOSS_HOME=/opt/jboss/wildfly \
+    JBOSS_HOME=/opt/wildfly \
```
ochaloup: @jmesnil Regarding what we talked at https://github.com/wildfly/wildfly-operator/pull/113. I think this should be changed to `/wildfly`. The WFLY17 and WFLY18 cent os images have a different location of the JBOSS_HOME files (ie. `/opt/jboss/wildfly` vs `/opt/wildfly`) but both images have a symlink to the location of `/wildfly`. If we want to get running the Operator with WFLY17 and WFLY18 without need to change the `JBOSS_HOME` property in the `operator.yaml` file I think it could be reasonable to use the `/wildfly` path.

jmesnil: That's a good idea.

I did not change it because `jboss-cli.sh` will complain when it is executed from the symlink and this makes the tx recovery command fails:

```
[jboss@d74328f5b39f ~]$ /opt/wildfly/bin/jboss-cli.sh -c
[standalone@localhost:9990 /]
[jboss@d74328f5b39f ~]$ /wildfly/bin/jboss-cli.sh -c
WARNING JBOSS_HOME may be pointing to a different installation - unpredictable results may occur.

[standalone@localhost:9990 /]
```

However now that #112 is fixed, this env var is only used to mount volumes and the symlink should work in both cases.

I'll update my PR.

ochaloup: Oh, I see, I didn't realize that. It's fine that's fine now :)

PR #107, opened 2019-10-07
S2I environmental values for EJB recovery would be setup

ochaloup:
The operator will provide the name of the headless service for the S2I script setting up the socket binding client mapping correctly.

### DIFF
--- a/pkg/resources/statefulsets/statefulset.go
+++ b/pkg/resources/statefulsets/statefulset.go
@@ -104,6 +104,9 @@ func NewStatefulSet(w *wildflyv1alpha1.WildFlyServer, labels map[string]string, 
 	// the application uses clustering and KUBE_PING.
 	statefulSet.Spec.Template.Spec.Containers[0].Env = append(statefulSet.Spec.Template.Spec.Containers[0].Env, envForClustering(k8slabels.SelectorFromSet(labels).String())...)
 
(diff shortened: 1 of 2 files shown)


pkg/resources/statefulsets/statefulset.go
```diff
@@ -290,3 +293,12 @@ func envForClustering(labels string) []corev1.EnvVar {
 		},
 	}
 }
+
+func envForEJBRecovery(w *wildflyv1alpha1.WildFlyServer) []corev1.EnvVar {
+	return []corev1.EnvVar{
+		{
+			Name:  "STATEFULSET_HEADLESS_SERVICE_NAME",
```
jmesnil: @ochaloup just a note that once an env var is used to setup a S2I module, this becomes an API and changing the name becomes more difficult (if only possible...).
Are you ok with that name or something more descriptive and specific to the tx recovery should be used?

ochaloup: My point was to use this generic name as it's reallythe name of the headless service which is managed by the `StatefulSet`. This name could be, potentially, used at other configuration too.

For our case the ENV property is used for socket binding setting correctly the client mapping and this configuration is needed for any ejb remote calls to the particular server.
From that I think it's ok this being named with this generic name not including any word like `recovery` or `transaction`.

PR #108, opened 2019-10-08
Adding socket dial dead timeout when recovery would be stuck

ochaloup:
Just a small fix on adding a timeout for a dead connection during recovery processing.

If the application server would be stuck for some reason it could let handing the connection. The dead timeout defines the time when recovery scan is considered as errorneous and the waiting will be released.

### DIFF
--- a/pkg/controller/util/remote_ops.go
+++ b/pkg/controller/util/remote_ops.go
@@ -28,7 +28,8 @@ var (
 	// start of the time timestamp
 	startOfTheTimeTimestamp = time.Unix(0, 0)
 	// socket dial timeout
-	socketDialTimeout = GetEnvAsDuration("SOCKET_DIAL_TIMEOUT", 10, time.Second)
+	socketDialTimeout = GetEnvAsDuration("SOCKET_DIAL_TIMEOUT", 30, time.Second)
+	socketDeadTimeout = GetEnvAsDuration("SOCKET_DEAD_TIMEOUT", 10*60, time.Second)
 )
 
 //ExecRemote executes a command inside the remote pod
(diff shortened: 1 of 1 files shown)


pkg/controller/util/remote_ops.go
```diff
@@ -28,7 +28,8 @@ var (
 	// start of the time timestamp
 	startOfTheTimeTimestamp = time.Unix(0, 0)
 	// socket dial timeout
-	socketDialTimeout = GetEnvAsDuration("SOCKET_DIAL_TIMEOUT", 10, time.Second)
+	socketDialTimeout = GetEnvAsDuration("SOCKET_DIAL_TIMEOUT", 30, time.Second)
+	socketDeadTimeout = GetEnvAsDuration("SOCKET_DEAD_TIMEOUT", 10*60, time.Second)
```
jmesnil: Is the value really 10 minutes?

ochaloup: it's something I'm not sure how to setup. I used a defensive approach and took a value that seems to be big enough.
The trouble is that the SCAN call is blocking - it's called and all recovery processing has to be done. It normally takes like a minute or two. But it's hard to expect how it behaves in case of some big cluster.

As a side note. Regarding the blocking approach - I have still in mind our conversation about better async/reactive approach. I've already created an issue for that (https://github.com/wildfly/wildfly-operator/issues/110) and with the gained experience I believe the next iteration of design of the scaledown recovery processing will be more "operator native".

jmesnil: ok, so in that case, could you update the commit to set it at `10, time.Minute` to make it explicit please?

ochaloup: The reason I used the `10*60 `is that the value is configurable by user. I expected the user will use seconds as they provide better granularity.
And the `SOCKET_DIAL_TIMEOUT` is in seconds then the `SOCKET_DEAD_TIMEOUT` would be good to be in seconds as well.

For sure I can change the method `GetEnvAsDuration` to work with values from env variable in seconds while the default value could be defined in a different unit. But it seems to me ambiguous.
Currently the time unit at `GetEnvAsDuration` defines the time unit for both - for the default value and for the value passed from the env variable.

Do you think a different approach should be used?

jmesnil: ok, that's fine

Issue #112, opened 2019-10-09
do not use resources.JBossHome for tx recovery scripts


jmesnil:
Transaction recovery scripts requires to execute CLI scripts by using the `jboss-cli.sh` console.

The path to the console is determined inside the application image where `JBOSS_HOME` is always set to the right location. There is no need to configure this path from the "outside" in the operator code.

PR #113, opened 2019-10-09
[#112] Use JBOSS_HOME to set the tx recovery script

jmesnil:
do not use resource.JBossHome to set the commands executed inside the
application image but rely on the presence of JBOSS_HOME instead

This fixes 112.


jmesnil:
@ochaloup fiy

pkremens:
@jmesnil could you please remove also the https://github.com/wildfly/wildfly-operator/blob/master/build/Dockerfile#L4 so it doesn't encourage anyone to use it

jmesnil:
@pkremens no, I still need to keep it to be able to mount volume at https://github.com/wildfly/wildfly-operator/blob/master/pkg/resources/statefulsets/statefulset.go#L183

(Note that this env var is internal to the operator image and only the code inside it can access it)

PR #127, opened 2019-11-22
[#126] disable transaction manager on scale down pod

### LINKED ISSUE #126: On server in shutdown state do not allow a new transaction to be started
When pod is in scaledown phase it should not be permitted for the container to start a new transaction.

This is a part of the solution for https://issues.jboss.org/browse/EAP7-1355

ochaloup:
Adding a fix for transaction manager processing being disabled during the scaledown process - aka. no new transaction can be started.
The recovery of existing transaction is still possible (verified with EAP QE testsuite, cc @simkam)

This fix is expected to be follow-up with a broader service stoping during the recovery in future.

fixes #126

### DIFF
--- a/pkg/controller/util/wildfly_mgmt.go
+++ b/pkg/controller/util/wildfly_mgmt.go
@@ -37,6 +37,8 @@ var (
 	MgmtOpSystemPropertyPeriodicRecoveryPeriod = "/system-property=com.arjuna.ats.arjuna.common.RecoveryEnvironmentBean.periodicRecoveryPeriod:add(value=%s)"
 	// MgmtOpSystemPropertyOrphanSafetyInterval is a JBoss CLI command to set system property of orphan safety interval
 	MgmtOpSystemPropertyOrphanSafetyInterval = "/system-property=com.arjuna.ats.jta.common.JTAEnvironmentBean.orphanSafetyInterval:add(value=%s)"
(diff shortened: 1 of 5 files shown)


jmesnil:
As a test, can you update the `.travis.yaml` file and set the Go version to `1.13.x` please?

jmesnil:
could you please rebase on top of  master and I'll merge it, thanks

ochaloup:
Oh, right. Done.
...plus I hope that the test will pass now. Thanks.

jmesnil:
@ochaloup failure is genuine, code does not compile:

```
# github.com/wildfly/wildfly-operator/pkg/resources/statefulsets
315pkg/resources/statefulsets/statefulset.go:152:14: undefined: resources.StandaloneServerDataDirPath
316FAIL	github.com/wildfly/wildfly-operator/pkg/controller/wildflyserver [build failed]
317
```

ochaloup:
hi @jmesnil . I'm a bit fighting with the operator testing for some time now. I tried on the local machine and I had got issues on WFLY17 for cluster tests. After some attempts of configuration of minikube it started to run. But now it fails with WFLY18 with a strange timeout error. This one can be seen here in the log too.
Do you have experience what that could be connected with? For my testing it seems it's not related to this PR, or?
Thanks!

ochaloup:
@jmesnil ok. I hope that I fixed the trouble (it seems to be my wrong perception). I checked with some of the eap integration tests. All seems to be fine. Do you think it could be merged now? :-p

jmesnil:
@ochaloup thanks

PR #130, opened 2019-12-20
[#27] Support ImageStream

### LINKED ISSUE #27: Update Operator's applicationImage when an ImageStream change is detected
To improve Seamless upgrade when the applicationImage is coming from OpenShift's ImageStream, the operator should detect changes to the image stream (e.g. when the sha of the tag changes) and uses the new sha to set its applicationImage.

The applicationImage from the spec should always refer to the image stream tag and an applicationImage in the operator status will correspond to the actual sha of the application image.

jmesnil:
* add Spec.SourceRepositorySpec to build the application image from a
  Git repository using WildFly S2I
* add Spec.BuildEnv to pass environment variables to the S2I build
* make Spec.ApplicationImage optional as the definition of
  SourceRepositorySpec allows an alternative
* Create ImageStream and BuildConfig resources to handle S2I builds
* Use image.openshift.io/triggers annotation to trigger statefulset
  update when the image stream changes
* add S2I_BUILDER_IMAGE env var to Dockerfile to specify which S2I
  builder image is used by the operator

This fixes #27
 Please enter the commit message for your changes. Lines starting


jmesnil:
superseded by #132

PR #131, opened 2019-12-20
updated README.adoc - replicas is a required spec

ckaserer:
as specified in the Custom Resource Definition

```
kind: CustomResourceDefinition
apiVersion: apiextensions.k8s.io/v1beta1
metadata:
  name: wildflyservers.wildfly.org
```

both `applicationImage` and `replicas` are required as specified here:

 ```
required:
        - applicationImage
        - replicas
```


jmesnil:
@ckaserer thanks

PR #139, opened 2020-03-04
[#138] Add permissions to be able to update the deployments/finalizers resource

### LINKED ISSUE #138: Error creating service object that exposes the operator metrics
When the operator is installed we can see the following message in the logs:

`{"level":"info","ts":1568012414.9460623,"logger":"cmd","msg":"failed to create or get service for metrics: services \"wildfly-operator\" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>"}`

It fails to create wildfly-operator service object which exposes the operator metrics.

The error is due to a lack of granted permissions for the operator. It needs to be able to set owner references to the metrics Service resource so that the metrics Service will get deleted as soon as the operator is deleted.

yersan:
This fixes #138

### DIFF
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -31,6 +31,7 @@ rules:
   - daemonsets
   - replicasets
   - statefulsets
+  - deployments/finalizers
   verbs:
   - '*'
 - apiGroups:


jmesnil:
@yersan thanks

PR #144, opened 2020-05-13
[#126] fix on wrong boolean handling, disable transaction manager on scale down

ochaloup:
fix for prior PR: https://github.com/wildfly/wildfly-operator/pull/127

I found I handled wrongly the boolean property as for disabling the transaction manager processing is needed to set to {{true}}.
The default is {{false}} - https://github.com/jbosstm/narayana/blob/5.10.4.Final/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/common/CoordinatorEnvironmentBean.java#L66


ochaloup:
@yersan I have no much idea what is wrong with this PR. I just need to rename a variable and change the boolean value.

yersan:
Yes @ochaloup , me neither. We have been getting some weird behaviour with the Ci lately, I don't have the details yet

simkam:
for some reason, CI doesn't report results back to PR. The same for #167, jobs finished (https://prow.ci.openshift.org/?repo=wildfly%2Fwildfly-operator) but results weren't updated on PR. I'll take a look on Monday.

simkam:
it looks like they again changed webhook URL. @jmesnil can you please try to update it - https://docs.ci.openshift.org/docs/how-tos/onboarding-a-new-component/#webhook-setup . I'm not sure whether it will help, because jobs start even with the current configuration, but at least it looks like a permanent URL now.

jmesnil (approved):

jmesnil:
CI jobs passed on https://prow.ci.openshift.org/pr-history/?org=wildfly&repo=wildfly-operator&pr=144

jmesnil (approved):

PR #145, opened 2020-05-28
Webhooks

jmesnil:


deploy/crds/conversion_webhook_template.yaml
```diff
@@ -0,0 +1,22 @@
+# apiextensions.k8s.io/v1beta1
+  preserveUnknownFields: false
+  conversion:
+    strategy: Webhook
+    conversionReviewVersions: ["v1","v1alpha1","v1alpha2"]
```
jmesnil, line 5: where does this "v1" version come from?

deploy/crds/conversion_webhook_template.yaml
```diff
@@ -0,0 +1,22 @@
...
+    webhookClientConfig:
+      service:
+        namespace: wildfly-op
```
jmesnil, line 8: can we remove this namespace?
Depending on how the operator is installed it might end up in different namespaces (we'll have to figure out for OLM)

deploy/operator.yaml
```diff
@@ -2,23 +2,31 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: wildfly-operator
+  labels:
+    name: wildfly-operator
+    app: wildfly-operator
```
jmesnil, line 7: did you add this label or was it autogenerated?

pkg/apis/wildfly/v1alpha1/wildflyserver_conversion.go
```diff
@@ -0,0 +1,129 @@
+package v1alpha1
```
jmesnil, line 1: Does the conversion has to be in the "old" version?
It would feel more natural that it goes into the "new" version( that knows the old package).

pkg/apis/wildfly/v1alpha1/wildflyserver_conversion.go
```diff
@@ -0,0 +1,129 @@
...
+			VolumeClaimTemplate: src.Spec.Storage.VolumeClaimTemplate,
+		}
+	}
+	dst.Spec.ServiceAccountName = src.Spec.ServiceAccountName
+	dst.Spec.EnvFrom = src.Spec.EnvFrom
+	dst.Spec.Env = src.Spec.Env
+	dst.Spec.Secrets = src.Spec.Secrets
+
+	// Status
+	dst.Status.Replicas = src.Status.Replicas
```
jmesnil, line 53: could you just to `dst.Status = src.Status`?

deploy/crds/wildfly.org_wildflyservers_crd.yaml
```diff
@@ -24,486 +24,1039 @@ spec:
...
+                  \n Read-only."
+                format: int32
+                type: integer
+            required:
+            - replicas
+            - scalingdownPods
+            type: object
+        type: object
+    served: true
+    storage: true
```
jmesnil, line 538: that does not seem right: I'd expect that the new `v1alpha2` would be the storage version instead.

pkg/apis/wildfly/v1alpha2/wildflyserver_types.go
```diff
@@ -0,0 +1,151 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// WildFlyServerSpec defines the desired state of WildFlyServer
+// +k8s:openapi-gen=true
+type WildFlyServerSpec struct {
```
jmesnil, line 13: We would have to find a convention in the comments to explains the difference to the previous version. It's difficult to figure that out when looking at the whole struct.

jmesnil: maybe there is an annotation similar to the javadoc `@since`

pkg/apis/wildfly/v1alpha2/wildflyserver_conversion.go
```diff
@@ -0,0 +1,4 @@
+package v1alpha2
+
+// Hub marks this type as a conversion hub.
+func (*WildFlyServer) Hub() {}
```
jmesnil, line 4: iirc, the Hub is in package `v1alpha2` so I'd expect that the corresponding converter method is in the same package

pkg/apis/wildfly/v1alpha1/wildflyserver_conversion.go
yersan: We convert from spoke to hub and from hub to another spoke. In our currently implementation, v1alpha2 is our Hub (the stored version in kubernetes) and v1Alpha1 is our spoke, currently there is only one spoke. This should clear up these concepts: https://book.kubebuilder.io/multiversion-tutorial/conversion-concepts.html

We have this conversion file here to implement the conversion, because we need to convert from v1alpha1, which implements the Convertible interface, to the type that implements Hub, in out case v1alpha2. Since we choose to store in Kubernetes the v1alpha2, we need to make v1alpha1 implement Convertible interface and then, add conversion methods here in the "old" version.

pkg/apis/wildfly/v1alpha1/wildflyserver_types.go
```diff
@@ -119,6 +120,7 @@ type PodStatus struct {
 // +kubebuilder:printcolumn:name="Replicas",type="integer",JSONPath=".spec.replicas"
 // +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
 // +kubebuilder:resource:shortName=wfly
+// +kubebuilder:storageversion
```
yersan, line 123: This is a mistake, the desired storage version is v1alpha2

pkg/apis/wildfly/v1alpha2/wildflyserver_conversion.go
yersan: In this case we selected v1alpha2 as the implementation of the Hub interface, so we do not need to add any conversion here.

deploy/crds/wildfly.org_wildflyservers_crd.yaml
yersan: You're right, a mistake, it is controlled by https://github.com/wildfly/wildfly-operator/pull/145/files#diff-553a69f3eee8fa0585c8bfe19a47a871R123

jmesnil:
Let's close this PR and reopen it when we will tackle the conversion webhook again

PR #151, opened 2020-07-29
add new target to Makefile for e2e tests in Prow (OpenShift CI)

simkam:
new target `test-e2e-prow` for e2e tests in Prow. I'll update CI configuration when this is merged.


Makefile
```diff
@@ -68,8 +68,10 @@ test-e2e-18-local: setup
 	LOCAL_OPERATOR=true JBOSS_HOME=/wildfly OPERATOR_NAME=wildfly-operator ./operator-sdk test local ./test/e2e/18.0 --verbose --debug  --namespace default --up-local
 
 ## test-e2e-18           Run e2e test for WildFly 18.0 with a containerized operator
+test-e2e-18: export component := wildfly-operator
```
jmesnil, line 71: for reference, could you please link to the CI job that uses that target?

I wonder if we should make a distinct target for it (so that the test-e2e-18 target can also be run locally by developers). something like `test-e2e-ci`?

jmesnil:
@simkam thanks!

PR #152, opened 2020-07-31
Adding support for IBM POWER(ppc64le)

bahetiamit:
Signed-off-by: Amit Baheti aramswar@in.ibm.com


build/setup-operator-sdk.sh
```diff
@@ -9,8 +9,11 @@ if ! [ -e "./operator-sdk" ]; then
     *)          file=operator-sdk-${RELEASE_VERSION}-x86_64-linux-gnu;;
   esac
 
+  ARCH='$(uname -m)'
```
jmesnil, line 12: should it be

```
ARCH=$(uname -m)
```
?

If I run `ARCH='$(uname -m)'` on my laptop, I have:
```
echo $ARCH
$(uname -m)
```

and not the actual arch of my laptop.

```
$ ARCH=$(uname -m)
$ echo $ARCH
x86_64
$ ARCH='$(uname -m)'
$ echo $ARCH
$(uname -m)
```

bahetiamit: @jmesnil You are right. The if condition in the script is always evaluating to true and hence was working fine when we tested on POWER machine. Also as many changes were made in the master branch, let me open a new clean PR.

bahetiamit:
Created https://github.com/wildfly/wildfly-operator/pull/155

PR #154, opened 2020-08-04
Skip setup target for test-e2e-prow

jmesnil:
as it will use a specific setup-operator-sdk-e2e-tests.sh to set another
version of the operator-sdk specific for e2e tests


simkam:
shouldn't it be skipped also for `test-e2e-18` which uses the same workaround?

PR #155, opened 2020-08-04
Adding support for IBM POWER(ppc64le)

bahetiamit:
Signed-off-by: Amit Baheti aramswar@in.ibm.com

### DIFF
--- a/build/setup-operator-sdk.sh
+++ b/build/setup-operator-sdk.sh
@@ -9,6 +9,9 @@ if ! [ -e "./operator-sdk" ]; then
     *)          file=operator-sdk-${RELEASE_VERSION}-x86_64-linux-gnu;;
   esac
 
+ARCH=$(uname -m)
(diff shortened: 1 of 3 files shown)


build/setup-operator-sdk.sh
```diff
@@ -9,8 +9,11 @@ if ! [ -e "./operator-sdk" ]; then
     *)          file=operator-sdk-${RELEASE_VERSION}-x86_64-linux-gnu;;
   esac
 
+ARCH=$(uname -m)
```
jmesnil, line 12: Could you please apply the same change to build/setup-operator-sdk-e2e-tests.sh?

We use a different version of the operator-sdk for the end-to-end tests but we need to make sure this version also uses the proper arch.

bahetiamit: Yes @jmesnil . I have made the required changes.

jmesnil:
@bahetiamit thanks

PR #164, opened 2020-10-29
[#25] Integrate with Prometheus metrics

### LINKED ISSUE #25: Expose WildFly Metrics to Prometheus by creating a ServiceMonitor
# Expose WildFly Metrics to Prometheus by creating a ServiceMonitor

## Issue Metadata

* https://github.com/wildfly/wildfly-operator/issues/25

## Related Issues

* [EAP7-1579 - As a developer, I want to monitor metrics from application deployed with the EAP Operator](https://issues.redhat.com/browse/EAP7-1579)

### Dev Contacts

* https://github.com/jmesnil

### QE Contacts

* [Jan Kasik](mailto:jkasik@redhat.com)

### Testing By

* [x] QE

### Affected Projects or Components

* [WildFly Operator](https://github.com/wildfly/wildfly-operator)

## Requirements

* Expose WildFly Metrics (from the `/metrics` HTTP endpoint on its management port) so that they can be monitored out-of-the-box by the OpenShift Monitoring tools (based on Prometheus). This is achieved by creating a `ServiceMonitor` resource (owned by the Prometheus Operator) that tells Prometheus servers that they can query metrics from WildFly on its management port.

### Non-Requirements

* The installation of the Prometheus operator (that provides the `ServiceMonitor`) is out of scope. If the WildFly Operator does not detect the presence on the `ServiceMonitor` CRD, it will do nothing.

### Implementation Plan

When the WildFly Operator watches a `WildFlyServer` resource, it will checks for the presence of the `ServiceMonitor` CRD and eventually create a `ServiceMonitor` that expose its metrics.
In any case the WildFly Operator also creates a `Service` to expose its admin/management port to the cluster.

## Community Documentation

This RFE does not impact the `WildFlyServer` CRD.
The community documentation will only mention that if the Prometheus Operator is deployed on the cluster (and the `ServiceMonitor` CRD is installed), then WildFly metric can be consumed out-of-the-box.

## Release Note Content

> Application deployed by the WildFly Operator  provides metrics that can be monitored out-of-the-box. It requires the installation of the Prometheus Operator.

jmesnil:
When ServiceMonitor resource is installed on the cluster,
create a ServiceMonitor to expose WildFly metrics.
The metrics are available on the management port of WildFly (9990)
using the standard Prometheus endpoints (/metrics).
A -admin service is created to expose the management port of WildFly and
is referenced by the ServiceMonitor resource.

This fixes #25.


pkg/controller/wildflyserver/wildflyserver_controller.go
```diff
@@ -223,6 +238,15 @@ func (r *ReconcileWildFlyServer) Reconcile(request reconcile.Request) (reconcile
 		}
 	}
 
+	// create a Prometheus ServiceMonitor (if the resource exists on the cluster)
+	if hasServiceMonitor() {
```
yersan, line 242: Should not be include the case when the ServiceMonitor resource was deleted on the cluster? If so, I would expect to delete any resource created in a previous loop.

In addition to that, is the service monitor resource created for WildFly deleted when the WildFly Operator is removed? ... I'm not sure if that is happening

jmesnil: The created ServiceMonitor is owned by the WildFlyServer resource. It will be deleted when the WildFlyServer resource itself is deleted.

If the ServiceMonitor CRD is removed from the cluster, the one created by the WildFlyServer will also be removed. When the reconciler loop is run, we will not create a new one as the CRD is no longer present.

Regarding the removal of the operator: when the operator is removed, it does not impact the resources that were deployed. The `WildFlyServer` CRD will still be available.
It's only when you delete the `WildFlyServer` CRD that you would delete the corresponding resource and any other resources that are owned by it (including its ServiceMonitor)

jmesnil:
The CI jobs are passing for this PR but somehow they have not reported back to GitHub:
https://prow.ci.openshift.org/pr-history/?org=wildfly&repo=wildfly-operator&pr=164

yersan (approved):

pkg/controller/wildflyserver/wildflyserver_controller.go
```diff
@@ -12,15 +12,19 @@ import (
 	wildflyutil "github.com/wildfly/wildfly-operator/pkg/controller/util"
 	"github.com/wildfly/wildfly-operator/pkg/resources"
 	"github.com/wildfly/wildfly-operator/pkg/resources/routes"
+	"github.com/wildfly/wildfly-operator/pkg/resources/servicemonitors"
 	"github.com/wildfly/wildfly-operator/pkg/resources/services"
 	"github.com/wildfly/wildfly-operator/pkg/resources/statefulsets"
 
+	monitoringv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
```
mchoma, line 19: I do not know go language. But does this mean wildfly operator is relying on life resource? If github will have outage operator will not work properly?

jmesnil: What do you mean by "life resources"?

This is equivalent to a Java package where we import the `monitoringv1` module from the GitHub repo when the application is built. It does not connect to github.com during the application runtime.

PR #165, opened 2020-11-09
Switch getPodStatus returns and improve function documentation

pprokopi:
I think it's clearer having the PodStatus array as the first return of getPodStatus.
Also added named returns to improve the function documentation.


yersan:
Hello @pprokopi , I'm closing this PR since it is an old one that got stale, sadly, we were not able to properly review it.

PR #183, opened 2021-02-03
Fix on sh scripts to not applying removed yaml files

ochaloup:
…where declaration was put under operator.yaml

As part of the commit for e2e tests for github some yaml declarations were moved under the `operator.yaml`. This should be reflected in the installation shell scripts.
https://github.com/wildfly/wildfly-operator/commit/ef0b424d05f9899e5e476367e2bd3c607c2ef173#diff-b7bd55c11b781b0ccc43aa6e57f9dadf0660e9d1d4e27e0979ee43a407d454ae

### DIFF
--- a/build/run-minikube.sh
+++ b/build/run-minikube.sh
@@ -1,9 +1,6 @@
 #!/usr/bin/env bash
 
-kubectl apply -f deploy/service_account.yaml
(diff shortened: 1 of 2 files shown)


yersan (approved):

yersan:
Thank you for spotting this @ochaloup

PR #185, opened 2021-02-08
Copy clusterIP when updating services

jmesnil:

### DIFF
--- a/pkg/resources/services/service.go
+++ b/pkg/resources/services/service.go
@@ -16,18 +16,18 @@ var log = logf.Log.WithName("wildflyserver_services")
 
 // CreateOrUpdateAdminService create a admin service or returns one up to date with the WildflyServer
 func CreateOrUpdateAdminService(w *wildflyv1alpha1.WildFlyServer, client client.Client, scheme *runtime.Scheme, labels map[string]string) (*corev1.Service, error) {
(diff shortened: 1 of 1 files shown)


jmesnil:
This fixes a regression introduced in #25 where that line was removed in https://github.com/wildfly/wildfly-operator/pull/164/files#diff-23e0afc6450903ba4ce6cd4ce4d7d537a69251a712402a5ad98b805c5eaec63fL81

jmesnil:
@ochaloup could you please review?

ochaloup (approved):
Oh, that seems nice. I'm stupid. When I saw the error I didn't recognize that it talks about this specific field (ie. `"error":"Service \"quickstart-loadbalancer\" is invalid: spec.clusterIP: Invalid value: \"\": field is immutable`) and from some reason I was thinking about the whole service.

My testing - on my CRC I run a quickstart and I work with scaledown&scaleup - shows this works nicely and fixes what's reported as an issue at https://issues.redhat.com/browse/JBEAP-20978?focusedCommentId=15751623#comment-15751623.
I can't confirm this fix has any other impact on the JBEAP issue - first I don't think so as currently the loadbalancer is recreated and operating (only slows down the processing), second I struggle with the EAP QE Testsuite to understand the errors and I need more time digging deeper.

Anyway I think this is demanded fix. I should ask you earlier about this error message. Thank you.

PR #186, opened 2021-02-11
Build examples workflow

yersan:
A new GitHub action to build the application images used by the examples on GitHub with the latest wildfly-s2i image.

The idea is wildfly-s2 will trigger this event, I'm going to add a PR there to do so. Alternatively, we can invoke manually the workflow by invoking /build/build-examples.sh

To make this work the following secrets have to be created on this repo:
secrets.QUAY_USERNAME: The wildfly-quickstarts quay.io user name
secrets.QUAY_PASSWORD: The wildfly-quickstarts quay.io password

These secrets will define the user/pass to push the built application images.

We also need a new access token with the following permissions:
repo:public_repo

This token will be used by the wildfly-s2 workflow to dispatch the event against this repo when a new wildfly-s2i image is published in quay.io. See: https://github.com/wildfly/wildfly-s2i/pull/143

I'm going also to add a follow-up PR to clean up the examples a bit and make them use the:latest tag as application images.

### DIFF
--- a/.github/workflows/build_examples.yml
+++ b/.github/workflows/build_examples.yml
@@ -1,0 +1,62 @@
+name: Wildfly Operator example images build
+
+on:
+  repository_dispatch:
+    types: [build_operator_examples]
+
+env:
+  LANG: en_US.UTF-8
+  S2I_URI: https://api.github.com/repos/openshift/source-to-image/releases/latest
+  QUAY_USERNAME: ${{ secrets.QUAY_USERNAME }}
+  QUAY_PASSWORD: ${{ secrets.QUAY_PASSWORD }}
+  EVENT_SOURCE: ${{ github.event.client_payload.source }}
+  EVENT_TYPE: ${{ github.event.action }}
+
+jobs:
+  build:
+    if: github.event.action == 'build_operator_examples'
+    name: Build Wildfly Operator examples
+    runs-on: ubuntu-latest
+    steps:
+    - name: Print info
+      run: echo "${EVENT_TYPE} received from ${EVENT_SOURCE}"
+    - name: Pull latest wildfly-s2i image
+      run: docker pull quay.io/wildfly/wildfly-centos7:latest
+    - name: Login quay.io
+      run: docker login -u="${QUAY_USERNAME}" -p="${QUAY_PASSWORD}" quay.io
+    - name: Install s2i binary
+      run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/build_examples.yml
```diff
@@ -0,0 +1,62 @@
...
+    name: Build Wildfly Operator examples
+    runs-on: ubuntu-latest
+    steps:
+    - name: Print info
+      run: echo "${EVENT_TYPE} received from ${EVENT_SOURCE}"
+    - name: Pull latest wildfly-s2i image
+      run: docker pull quay.io/wildfly/wildfly-centos7:latest
+    - name: Login quay.io
+      run: docker login -u="${QUAY_USERNAME}" -p="${QUAY_PASSWORD}" quay.io
+    - name: Install s2i binary
```
jmesnil, line 27: This might be simplified using https://github.com/redhat-actions/s2i-build. Have you tried it?

.github/workflows/build_examples.yml
```diff
@@ -0,0 +1,62 @@
+name: Wildfly Operator example images build
+
+on:
+  repository_dispatch:
+    types: [build_operator_examples]
+
+env:
+  LANG: en_US.UTF-8
+  S2I_URI: https://api.github.com/repos/openshift/source-to-image/releases/latest
+  QUAY_USERNAME: ${{ secrets.QUAY_USERNAME }}
```
jmesnil, line 10: Please use:

```
QUAY_USERNAME: wildfly-quickstarts+github_actions
QUAY_PASSWORD: ${{ secrets. QUICKSTARTS_ROBOT_PASSWORD }}
```

jmesnil (changes requested):

.github/workflows/build_examples.yml
jmesnil: I looked at it and it's not usable: it does not allow to specify a Git ref (it's expected to build the checked out repo which is not our case here)

jmesnil (changes requested):

.github/workflows/build_examples.yml
yersan: Thanks for let me know, I was trying it just now :)

.github/workflows/build_examples.yml
yersan: Done

yersan:
We are going to move this workflow to wildfly-s2i, so, closing this

PR #192, opened 2021-04-01
 [#191] fix for scaling up during recovery processing

### LINKED ISSUE #191: Pod which is scaled up during transaction recovery processing could not be fully activated
There could be a trouble with a pod which is asked for being scaled down. When it's in middle of the transaction recovery processing there are applied changes on annotations and labels. These are pointers used by transaction recovery processing to first deactivate the pod from the processing new requests (to finish only what is un-finished) second to annotate such pod with information on the stage of the recovery processing (the reconciliation happens several times and the process of recovery tries to start at place where it has finished).

When the WildFly application is asked to scale up again the pod in the middle of the recovery processing could not be cleaned from all the pointers.
The pod scaling down pod needs to be refreshed as a clean one when it's made active.

ochaloup:
fixes #191

### DIFF
--- a/pkg/controller/wildflyserver/wildflyserver_controller.go
+++ b/pkg/controller/wildflyserver/wildflyserver_controller.go
@@ -2,12 +2,12 @@ package wildflyserver
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"reflect"
 	"strconv"
 	"strings"
-	"errors"
 
 	wildflyv1alpha1 "github.com/wildfly/wildfly-operator/pkg/apis/wildfly/v1alpha1"
 	wildflyutil "github.com/wildfly/wildfly-operator/pkg/controller/util"
@@ -139,7 +139,7 @@ func (r *ReconcileWildFlyServer) Reconcile(request reconcile.Request) (reconcile
 	}
 
 	if ok, err := validate(wildflyServer); !ok {
-		return r.ManageError(wildflyServer, err)
+		return r.manageError(wildflyServer, err)
 	}
 
 	// If statefulset was deleted during processing recovery scaledown the number of replicas in WildflyServer spec
@@ -434,13 +434,13 @@ func (r *ReconcileWildFlyServer) checkStatefulSet(wildflyServer *wildflyv1alpha1
 	return requeue, nil
 }
 
-func (r *ReconcileWildFlyServer) ManageError(w *wildflyv1alpha1.WildFlyServer, err error) (reconcile.Result, error) {
+func (r *ReconcileWildFlyServer) manageError(w *wildflyv1alpha1.WildFlyServer, err error) (reconcile.Result, error) {
 	if err == nil {
 		r.recorder.Event(w, v1.EventTypeWarning, "WildFlyProcessingError", "Unknown Error")
 		return reconcile.Result{}, err
 	}
 
-	r.recorder.Event(w, v1.EventTypeWarning, "WildFlyProcessingError",err.Error())
+	r.recorder.Event(w, v1.EventTypeWarning, "WildFlyProcessingError", err.Error())
 	return reconcile.Result{}, err
 }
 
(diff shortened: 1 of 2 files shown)


pkg/controller/wildflyserver/wildflyserver_controller.go
```diff
@@ -139,7 +139,7 @@ func (r *ReconcileWildFlyServer) Reconcile(request reconcile.Request) (reconcile
 	}
 
 	if ok, err := validate(wildflyServer); !ok {
-		return r.ManageError(wildflyServer, err)
+		return r.manageError(wildflyServer, err)
```
yersan, line 142: Thanks for this. The idea was to use this method from the outside by using a validation webhook, but sounds fine to reduce the scope now.

ochaloup: Oh, I didn't know the context that the method should be used outside. I just checked where it's used and it made sense to me make it private. If you say I will change it back.

pkg/controller/wildflyserver/wildflyserver_controller.go
```diff
@@ -279,12 +279,18 @@ func (r *ReconcileWildFlyServer) Reconcile(request reconcile.Request) (reconcile
...
 	if numberOfPodsToScaleDown <= 0 {
 		for k, v := range wildflyServer.Status.Pods {
-			if v.State != wildflyv1alpha1.PodStateActive {
+			if v.State != wildflyv1alpha1.PodStateActive { // was in scale down processing
 				wildflyServer.Status.Pods[k].State = wildflyv1alpha1.PodStateActive
 				updateWildflyServer = true
+				// the non-active pod may be in the middle of the scale down process, we need to completely refresh it
+				for _, podItem := range podList.Items {
+					if podItem.Name == v.Name {
+						resources.Delete(wildflyServer, r.client, &podItem)
```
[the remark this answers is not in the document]
ochaloup: It's a good point just I think it's not appropriate to change the `podList` here. The list of pods are those which are known to `WildFlyServer` instance. That list is the same even the `Delete` is called. The delete only removes the `Pod` instance but `StatefulSet` immediately starts it back. The list of pod which is expected to be active for `WildFlyServer` remains unchanged. The `Delete` works as a way to restart/refrest the `Pod` instance setup - when recovery was in processing then the container setup was changed  - it could be in different ways depending what stage of the processing was hit by scale up.
Is the explanation of the purpose enough to have it as it is or did I miss some point here?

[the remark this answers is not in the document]
yersan: Thank @ochaloup , we should be fine then.

I was thinking about the possibility when the `StatefulSet` immediately doesn't start the pod, in that case, the `WildFlyServer` status will continue using obsolete information, for example, the pod IP. If the pod is deleted, its IP will change. 

In any case, the `podList` is read again at the beginning of the reconcile loop, if the pending pod is not started yet, the loop will wait at the beginning, and once it is available the `podList` will contain the correct data, which means rest of the code will be using real information.

`WildFlyServer` status will be updated at the end of the loop with the correct information again. So, it would be a corner case where the `WildFlyServer` status could take more time to show the real information. It should not be a problem.

pkg/controller/wildflyserver/wildflyserver_controller.go
yersan: No, it is fine as it is, thanks!

yersan (approved):

pkg/controller/wildflyserver/wildflyserver_controller.go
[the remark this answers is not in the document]
ochaloup: Yes, the WildFly server is always "eventually consistent" - e.g. pod could be just deleted manually by a user.

pkg/controller/wildflyserver/wildflyserver_controller.go
ochaloup: Thanks.

yersan:
Thank you @ochaloup

PR #194, opened 2021-04-01
[Issue#193] refresh pod when failure on transaction recovery restart happens

### LINKED ISSUE #193: Recovery processing requires container to be restarted but does not refresh pod when an error happens
The current implementation of the scale down recovery processing requires container to be restarted (this is considered as a deficiency an needs to be fixed via https://issues.redhat.com/browse/WFLY-12610 while restart can't be used for https://issues.redhat.com/browse/JBEAP-20714 working).

The restart of the container could not finish correctly and in such case the recovery processing could not be finished correctly. In case of the restart error there should be the pod taken down and launched again to run whole recovery process from scratch.

ochaloup:
fixes #193 

The fix makes the pod to consider to be refreshed when recovery container restart fails to proceed.


pkg/controller/wildflyserver/transaction_recovery.go
```diff
@@ -179,29 +188,35 @@ func (r *ReconcileWildFlyServer) setupRecoveryPropertiesAndRestart(reqLogger log
...
 
+		reqLogger.Info("Setting system property 'org.wildfly.internal.cli.boot.hook.marker.dir' at '/tmp/markerdir/wf-cli-shutdown-initiated'", "Pod Name", scaleDownPodName)
+		wildflyutil.ExecRemote(scaleDownPod, "mkdir /tmp/markerdir && touch /tmp/markerdir/wf-cli-shutdown-initiated || true")
+		wildflyutil.ExecuteMgmtOp(scaleDownPod, "/system-property=org.wildfly.internal.cli.boot.hook.marker.dir:add(value=/tmp/markerdir)")
+
+		reqLogger.Info("Restarting application server to apply the env properties", "Pod Name", scaleDownPodName)
+		if err := wildflyutil.ExecuteOpAndWaitForServerBeingReady(reqLogger, wildflyutil.MgmtOpRestart, scaleDownPod); err != nil {
+			reqLogger.Error(err, "Cannot restart application server after setting up the periodic recovery properties, "+
+				"pod will be deleted to be relaunched.", "Pod Name", scaleDownPodName)
+			resources.Delete(w, r.client, scaleDownPod)
```
yersan, line 204: Hello @ochaloup , here we are deleting the pod in the middle of a transaction recovery process after getting an error. The pod will be deleted and recreated again. What happens with the transaction state if we remove the pod? Are not all the transaction state missing at this point?

I also noticed on L:211 there is a try to update the pod which has been deleted on this line. Is that correct and expected?

pkg/controller/util/wildfly_mgmt.go
```diff
@@ -151,28 +151,47 @@ func GetTransactionRecoveryPort(pod *corev1.Pod) (int32, error) {
 // ExecuteOpAndWaitForServerBeingReady executes WildFly management operation on the pod
 //  this operation is checked to succeed and then waits for the container is ready
 //  this method is assumed to be used for reload/restart operations
-func ExecuteOpAndWaitForServerBeingReady(reqLogger logr.Logger, mgmtOp string, pod *corev1.Pod) (bool, error) {
+//  returns error if execution was not processed succesfully
```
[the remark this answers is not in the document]
ochaloup: +1

pkg/controller/wildflyserver/transaction_recovery.go
ochaloup: Thanks @yersan for looking into this. I was just thinking to ask you for  review. I wanted to do so last week but I would like to verify these changes with help the EAP QE OpenShift testsuite and there was some technical issues that I was not able to do so. I will run the tests today and then report back to you here.

To your questions.
Deleting the pod is not a trouble. In fact it could happen that some user will do `oc delete pod transactional-application-0` anytime during the app processing. Transactions are not lost as the whole operator processing uses the statefulset. The transaction log is saved at the location shared amongst all executions of the pod of the same name. I do deliberately delete the pod to be restarted, launched with the default configuration and then processed by recovery handling from scratch. But the transaction log is available for the relaunched pod as it was for the prior one.

Thanks a lot for the second point with `L:211`. I was testing the behaviour on my CRC but it seems I missed this. The pod after deleting should not be updated for sure. During my testing the pod was marked for deletion and meanwhile it was updated and then failed. It was not an issue in processing (at least I did not hit it) but it is wrong. I will fix it.

yersan: In the aim to get a better understanding about it, some questions:

1. Where are the transaction recovery logs / entries stored by default?
2. If the transaction recovery process fails always without a recoverable way, is there any way to force the stateful set replica to scale down?

ochaloup: ad 1) the transaction recovery logs are placed under `data`  directory[1] (by default, it could be changed by user or with env variables, an option is to be saved into database)

ad 2) currently there is no direct way to force the scale down. we have discussed that during the development of the feature and there was some pushback about safety from my collegues
later I agreed with QE that such feature would be needed, the issue is logged at https://issues.redhat.com/browse/EAP7-1394 as RFE. I'm thinking to look at it but I haven't got time to start working on it.

[1]
```
...
                f:volumeMounts:
                  .: {}
                  k:{"mountPath":"/opt/eap/standalone/data"}:
                    .: {}
                    f:mountPath: {}
                    f:name: {}
...
```

yersan: @ochaloup thanks for the update.

About 1) Notice that by default, the operator data dir is backed by an EmptyDir volume type. Such a volume type does not survive a pod deletion.
About 2) Ok

Another detail to take into consideration for this feature is we have to verify if transactions (and all the stuff needed by the recovery process) is available on the application image as a safety check. If it is not available, then we could skip all the recovery proccess.

ochaloup: ad 1) it's a good point about the `EmptyDir` volume type but it's the default volume type for the Pod. The persistency should be ensured by `StatefulSet` (https://kubernetes.io/docs/tutorials/stateful-application/basic-stateful-set/#writing-to-stable-storage). I beleive the `data` directory is persisted over the `Pod` restarts.

ad safety check) ah, that's a good point. I'm just thinking on how to do so easily. Would you have a tip for me? Do you think like checking existence of the `transaction` subsystem? I think I need to create a new issue on this here when I'm clear if I got right what you mean.

yersan: > volume type but it's the default volume type for the Pod

Yes, and that's the issue. We specify it on the Documentation: https://github.com/wildfly/wildfly-operator/blob/master/doc/user-guide.adoc#storage. "The application may require persistent storage for some data (e.g. transaction or messaging logs) that must persist across Pod restarts."

Since we have it somehow documented, then we can assume it is on the user side to use a nondefault volume if transaction recovery is a requirement. That would be enough. However, with the new changes introduced in this PR, we could have now the situation where the pod can be restarted in the middle of a transaction recovery process, which means, under a default configuration, the transaction would never be recovered but the scale down will finish correctly being completely transparent for the user. In my opinion, if we are using the default configuration, we should not kick off the transaction recovery because there is no guarantee it will work as expected now. We could allow it, but if all the processes finish correctly, the user would think the transaction could have been recovered, but there is no guarantee about that situation.

> Would you have a tip for me? Do you think like checking existence of the transaction subsystem? 

Yes, I suggest a simple read-resource-operation to verify if the transaction (and all the subsystems involved on the transaction recovery process) are in place before starting the transaction recovery, if they are not, then avoid it, and inform the user.

ochaloup: ad storage) I think we are not talking about the same thing. The Pod's storage is `EmptyDir` but the `StatefulSet` should ensure to mount the persitent volume for data under the `data` directory. That's the place where txn log is store.
On top of that I can say we have some QE tests which does some failure recovery stuff (not sure exactly what but killing Pod should be part of it) and the txn log was not lost. I remember the same from my testing in prior days.
If you have a different experience then I will need to test it. If the `StatefulSet` persistency guarantees (see https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#using-statefulsets) does not hold true then all the recovery stuff handling needs to be rethought. Which I believe is not the case here.

And I really believe this is not a correct statement

> but there is no guarantee about that situation.

ad transaction subsystem) ok, I will create an issue on that. (a note: the recovery process is handled by transaction subsystem itself)

ochaloup: I created the issue here - https://github.com/wildfly/wildfly-operator/issues/195
I will look at it when we finish the discussion and adjustment on this one.

yersan: > The Pod's storage is EmptyDir but the StatefulSet should ensure to mount the persitent volume for data under the data directory. 

The volume is mounted on the data directory, correct, but what is mounted is an [EmptyDir](https://kubernetes.io/docs/concepts/storage/volumes/#emptydir) volume type by default. This is an ephemeral volume type that does not survive a pod deletion, it survives a pod crash though but not a pod deletion. We have it documented [here](https://kubernetes.io/docs/tutorials/stateful-application/basic-stateful-set/#writing-to-stable-storage) as well. That's the reason I don't believe delete the pod in the middle of recovery is fine.

The verification is simple, create a simple deployment with the Operator, create a file on the data directory on one pod, then delete the pod. The file created will be lost.

ochaloup: @yersan first of all, thanks and sorry.

I've got to run investigation on what I understood and clearly you are right. I'm sorry for being dumb in it. I just remembered the days when I was working with the statefulset and running the QE tests which are passing.
I'm now clear that StatefulSet has to be configured with PVC (which I did in past, which QE tests has got configured as it seems the Operator from OLM configures it). When `EmtpyDir` is used as the storage then the data is, logically, removed on Pod restart.

But from here I would need to clarify the point. The transaction recovery does not work with `EmptyDir` being set as the backend storage. There is no guarantee that the Pod is not rescheduled, a new env variable is defined or killed by user `oc` command. When the txn log does not survive such situation then it's not safe.

Operation on killing the pod assumes that the txn processing is set to be safe. When I remove the command restarting the pod from this PR I won't make the processing more safety when `EmptyDir` is used.

I agree with point that restarting the container would make sense for my situation. But as far as I can say there is no command which makes possible to just restart the container. Or would you have an idea how to proceed? (I'm thinking the `docker kill` is not a good option but maybe I'm wrong in such assumption.)

Or maybe put it differently. Now I do understand you are completely right about Pod restart. What do you propose to do to fix my wrong assumption which is manifested in the code of this PR with deleting the pod?

yersan:
Don't worry @ochaloup , I arrived recently at these concepts too and realize the situation looking at your PR.

I don't have any suggestions about how to properly kill the container instead of the pod. 

> Would be documenting - mentioning in doc + emitting a message in Operator log - plus, the recovery processing could be omitted as you suggested. Would that be what you think is needed?

I think so. If there is no guarantee that the transaction recovery works by using EmptyDir, we should just skip the transaction recovery facility, maybe emit an event or log trace, and allow the pod to be scaled down immediately. We have to be more explicit about this situation describing it in the Operator documentation. For example, the transaction recovery facility is only enabled if the CR if the storage is configured by using a PVC template, if not, transaction recovery is skipped. That's my opinion.

ochaloup:
@yersan I do experiments here with the settings here and I recognized that skipping the recovery when `EmptyDir` is used is not right. The transaction manager can be storing data in database and then `EmptyDir`does not harm the txn log persistency over pod restarts.

* With that I decided that I don't want to skip recovery for `EmptyDir` being used as the server configuration could be various.
* I updated the `user-guide.doc` with a warning talking about this point. 
* I think I need to address the point about skipping recovery - i.e. https://issues.redhat.com/browse/EAP7-1394 - as soon as possible.

Would you be fine to merge this PR with such statement? Do you think differently?

yersan:
The benefits I see if we are able to detect the transaction recovery can be automatically disabled, for example in the case of EmptyDir plus `/subsystem=transactions:read-attribute(name=use-jdbc-store)` returning false, is the pods will be scaled down quickly and with a less probability of a failure that can make the scale down process not possible.

At first glance, detecting the EmptyDir+reading use-jdbc-store sounds feasible to me, but I leave the decision up to you @ochaloup. If you think that would not be enough or you foresee other difficulties, we should be fine with changes incorporated by this PR.

Do you prefer to defer disabling the recovery in the future by using EAP7-1394 or do you want to disable it automatically now when possible by EmptyDir+reading use-jdbc-store case?

ochaloup:
@yersan ok, I understand your preference. I will try to add the handling you're talking about in update of this PR.

I consider to work on EAP7-1394 as soon as possible to address the deficiency of not being able to disable scale-down recovery forcibly.

pkg/controller/wildflyserver/transaction_recovery.go
jmesnil: @ochaloup Could you please provide some design about why deleting a pod is necessary?

This is an anti pattern and I don't see any good reason to allow it.
I'd prefer we focus energy on fixing the actual issue (that the tx subsystem requires a restart before doing recovery) and do not have to do such hacks in the operator to workaround this anti pattern

ochaloup: @jmesnil in the PR I'm fixing the behavior to work with the current code. I work on the fix for Narayana/WildFly to provide a way for not forcing the recovery but it's not available in current or previous releases of WildFly.

The motivation for restarting the pod is having a safety net for the deadlock behaviour described at https://issues.redhat.com/browse/WFCORE-5380. The restart of the app server is currently needed and there is a chance the cli script locks the processing. This does not happen normally this recovers when such an occurrence would happen.
If you consider such handling is malicious I will remove it. Just, please, confirm it here once again. The PR is prepared and it fixes few issues and I would like to promote it for merging.

I would like to learn why deleting of the the pod is an anti-pattern. Can you share with me some documentation or design article for me to understand the practices to follow?

jmesnil: @ochaloup the current code is convoluted and I'm not sure to understand the intent: during scaledown (so when a pod is about to be deleted), we stop the server, modify its configuration to enable recovery, and restart it. If the restart fails, we delete the pod (that is scheduled to be deleted!) to recreate it to delete it!

at that point, why do we bother with restarting the server inside the pod and can we not just delete the pod so that it is created again and always restarted (so that it can be properly deleted once tx recovery is achieved)?

 Do you see why I am confused?

ochaloup: @jmesnil I see I'm sorry that I was not able to communicate my intent better.
Maybe I should add some more descriptive comment to the code? Or maybe to write a paragraph on scale down process to README?

About the exact intention for this particular change. The trouble of the WFCORE-5380 is that on container restart the container can be stuck and the liveness check is not capable to delete it. The liveness goes to the `kill SIGTERM` which ends up at the not responding container. The killing pod tries with `SIGTERM` and then after the graceful shutdown time it calls the `SIGKILL`.

The process of the recovery works like: first it sets up the necessary properties (the recovery timeouts, txn enablement etc.), second restart the application server, third process recovery, fourth change the statefulset definition to scale down the number of pods.
This code change works at the second "step". The operator tries to restart the application server, if that attempt fails it force the pod deletion but it does not change the statefulset definition. Because of that the statefulset immediately restarts the pod and the next reconcile cycle tries to setup the recovery once again from scratch. The idea is that the deadlock is a race condition that will not be fulfilled next time and the next attempt would be hopefully successful.
Does this make better sense? I would be glad if you have an idea for me to improve my processing.

jmesnil: @ochaloup thanks for the explanation. let's fix the current behaviour to avoid such issues. 

But once we have the Narayana/WildFly fix, a lot of this code will be able to be removed.

yersan (approved):

PR #198, opened 2021-04-29
Add labels to the statefulset's template

### LINKED ISSUE #197: add metering labels to the pods managed by the operator
The downstream EAP Operator needs to add metering labels to the Pods that are managed by the operator to be able to identify and collect information on its workload.

We can add a env var named `POD_LABELS` that would contain a JSON string of labels key/values.
These labels would be added to the statefulset template in https://github.com/wildfly/wildfly-operator/blob/95a2ba43c62c2cfcda293baf0b4cb01c13642dce/pkg/resources/statefulsets/statefulset.go#L79 so that they are applied to any pods created by this statefulset

jmesnil:
If the STATEFULSET_TEMPLATE_LABELS env var contains a JSON
struct, the key and values are added to the labels of the
statefulset's podTemplate

This fixes #197

### DIFF
--- a/pkg/resources/constants.go
+++ b/pkg/resources/constants.go
@@ -25,6 +25,8 @@ const (
 	SecretsDir = "/etc/secrets/"
 	// ConfigMapsDir is the the directory to mount volumes from ConfigMaps
 	ConfigMapsDir = "/etc/configmaps/"
+	// StatefuleSetTemplateLabelsEnvVarName is the name of the envvar containg label/value map for pods created from the statefulset's template
+	StatefuleSetTemplateLabelsEnvVarName string = "STATEFULSET_TEMPLATE_LABELS"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


yersan:
I think using specific fields instead of pure JSON structure on an environment variable would scale better. Something like:

```
spec:
...
   podMetadata:
      labels:
         my/custom/label: my-value
```

It looks more natural to me and can evolve to add other pod metadata in the future, for example, annotations. I'm not sure if that would fit your needs @jmesnil

jmesnil:
@yersan these labels will be injected from an environment variable provided by the container image. Using a JSON structure allows to pass them as a single string value.

The only requirement for this feature is to be able to provide labels specifics to the EAP Operator with:

```
STATEFULSET_TEMPLATE_LABELS={\"com.company\": \"Red_Hat\", \"com.redhat.product-name\": \"Red_Hat_Runtimes\", \"com.redhat.product-version\": \"2021-Q2\", \"com.redhat.component-name\": \"EAP\", \"com.redhat.component-type\": \"application\"}<br class="Apple-interchange-newline"
```

yersan:
Ok, I see @jmesnil, I thought we wanted to expose this capability to the user by using the CR.

yersan (approved):
LGTM, just found a minor. It should not prevent merge it

Issue #208, opened 2021-09-21
No way to control cpu and memory resources from top level Wildfly YAML


seeker25:
Currently it defaults what the namespace defaults are. I believe this is under control of a cluster admin? 

`$ oc describe limits default-limits
Name:       default-limits
Namespace:  xxxxx
Type        Resource  Min  Max  Default Request  Default Limit  Max Limit/Request Ratio
----        --------  ---  ---  ---------------  -------------  -----------------------
Container   cpu       -    -    50m              250m           -
Container   memory    -    -    256Mi            1Gi            -`

As a work around I've had to do an OC patch to set the StatefulSet resources:

`oc patch statefulset xxx -p '{"spec": {"template": {"spec": {"containers": [{"name": "xxx", "resources": {"limits": {"cpu": "1000m","memory": "1Gi"}}}]}}}}'`

I figured the storage is already on there, why not have the CPU and memory as well for each replica?

https://access.redhat.com/documentation/en-us/red_hat_jboss_enterprise_application_platform/7.3/html-single/getting_started_with_jboss_eap_for_openshift_container_platform/index#wildflyserverspec_default

Wouldn't it be possible to pass down from WildFlySpec / WildFlyServerSpec to the StatefulSet here?:

https://github.com/wildfly/wildfly-operator/blob/6aba9ae99cd4324ba5e6f1e2501a65c52e1ec856/pkg/resources/statefulsets/statefulset.go#L87

danbigthink:
> Currently it defaults to what the namespace defaults are. I believe this is under control of a cluster admin?
> 
> $ oc describe limits default-limits
> Name: default-limits
> Namespace: xxxxx
> Type Resource Min Max Default Request Default Limit Max Limit/Request Ratio
> 
> Container cpu - - 50m 250m -
> Container memory - - 256Mi 1Gi -
> 
> As a work around I've had to do an OC patch to set the StatefulSet resources:
> 
> `oc patch statefulset xxx -p '{"spec": {"template": {"spec": {"containers": [{"name": "xxx", "resources": {"limits": {"cpu": "1000m","memory": "1Gi"}}}]}}}}'`
> 
> I figured the storage is already on there, why not have the CPU and memory as well for each replica?
> 
> https://access.redhat.com/documentation/en-us/red_hat_jboss_enterprise_application_platform/7.3/html-single/getting_started_with_jboss_eap_for_openshift_container_platform/index#wildflyserverspec_default
> 
[...]

PR #210, opened 2021-12-09
Allow autoscaling the operator with HPA

TomerFi:
**Added a [Resources](https://pkg.go.dev/k8s.io/api@v0.18.14/core/v1#ResourceRequirements) field to the `WildFlyServerSpec` type.**

This will allow the user to specify the CPU requirement for containers,
which is mandatory for the HPA calculation of CPU utilization.

Example of HPA eligible application:

```yaml
kind: WildFlyServer
metadata:
  name: quickstart
spec:
  applicationImage: "quay.io/wildfly-quickstarts/wildfly-operator-quickstart:18.0"
  replicas: 2
  resources:
    requests:
      cpu: "0.5"
```

---------------------------------------------------

**Added a string `Selector` field of type string to the `WildFlyServerStatus` type.**

This will be used for storing the selector label which is populated by the operator with the key [app.kubernetes.io/instance](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels).
The value for the label is created from the prefix `wildfly-` concatenated with the application name.
The same label is also applied to the underlying containers via the StatefulSet's template.

This is required for targeting the CR with the HPA instead of the underlying StatefulSet,
which is required to maintain the CR as the source for any of its resources.

### DIFF
--- a/pkg/controller/wildflyserver/wildflyserver_controller.go
+++ b/pkg/controller/wildflyserver/wildflyserver_controller.go
@@ -145,6 +145,9 @@ func (r *ReconcileWildFlyServer) Reconcile(request reconcile.Request) (reconcile
 		return r.manageError(wildflyServer, err)
 	}
 
+	// Set the selector label, this should be done for the pods as well to allow targeting the CR with HPA
+	wildflyServer.Status.Selector = fmt.Sprintf("app.kubernetes.io/instance=wildfly-%s", wildflyServer.GetName())
+
 	// If statefulset was deleted during processing recovery scaledown the number of replicas in WildflyServer spec
 	//  does not defines the number of pods which should be left active until recovered
 	desiredReplicaSizeForNewStatefulSet := wildflyServer.Spec.Replicas + wildflyServer.Status.ScalingdownPods
@@ -539,6 +542,7 @@ func LabelsForWildFly(w *wildflyv1alpha1.WildFlyServer) map[string]string {
 	labels["app.kubernetes.io/name"] = w.Name
 	labels["app.kubernetes.io/managed-by"] = os.Getenv("LABEL_APP_MANAGED_BY")
 	labels["app.openshift.io/runtime"] = os.Getenv("LABEL_APP_RUNTIME")
+	labels["app.kubernetes.io/instance"] = fmt.Sprintf("wildfly-%s", w.GetName())
 	if w.Labels != nil {
 		for labelKey, labelValue := range w.Labels {
 			labels[labelKey] = labelValue
--- a/deploy/crds/quickstart-resources-cr.yaml
+++ b/deploy/crds/quickstart-resources-cr.yaml
@@ -1,0 +1,14 @@
+apiVersion: wildfly.org/v1alpha1
+kind: WildFlyServer
+metadata:
+  name: quickstart
+spec:
+  applicationImage: "quay.io/wildfly-quickstarts/wildfly-operator-quickstart:18.0"
+  replicas: 2
+  resources:
+    limits:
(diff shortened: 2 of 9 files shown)


pkg/controller/wildflyserver/wildflyserver_controller.go
```diff
@@ -539,6 +542,7 @@ func LabelsForWildFly(w *wildflyv1alpha1.WildFlyServer) map[string]string {
 	labels["app.kubernetes.io/name"] = w.Name
 	labels["app.kubernetes.io/managed-by"] = os.Getenv("LABEL_APP_MANAGED_BY")
 	labels["app.openshift.io/runtime"] = os.Getenv("LABEL_APP_RUNTIME")
+	labels["app.kubernetes.io/instance"] = fmt.Sprintf("wildfly-%s", w.GetName())
```
jmesnil, line 545: I'm a bit confusing with the recommendations for these labels (looking at https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/).

* app.kubernetes.io/name -> The name of the application (eg `mysql`)
* app.kubernetes.io/instance -> A unique name identifying the instance of an application (e.g. `mysql-abcxzy`)

The key difference being that an application `name` can be installed many times in the same cluster (or namespace) while an `instance` is unique.

The existing `app.kubernetes.io/name` matches the CRD name (`w.Name` which is the same that `w.GetName()` and is thus already unique in the namespace (which is a stronger semantic that the recommendations from kubernetes.io). We could use the same value for the  `app.kubernetes.io/instance` without losing any information.

jmesnil (changes requested):
Aside from my comment on the label, this looks good.

Thanks a lot for fixing this issue!

pkg/controller/wildflyserver/wildflyserver_controller.go
TomerFi: Hmm... yes that is confusing, I haven't noticed that.

This explains that an application can have multiple instances on the same namespace:
https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#applications-and-instances-of-applications

That got me thinking,
maybe we shouldn't use the `app.kubernetes.io/instance` label,
or any other recommended label for that manner, as the selector label.
(maybe we should use the `instance` label regardless of the selector label, but that's for another discussion). 

To make stuff more robust, maybe we should use our own custom label,
something in the sorts of:

```go
labels[fmt.Sprintf("%s/%s", os.Getenv("LABEL_APP_RUNTIME"), "selector")] = fmt.Sprintf("%s-%s", os.Getenv("LABEL_APP_RUNTIME"), w.Name)
// or simply
labels[fmt.Sprintf("%s/%s", os.Getenv("LABEL_APP_RUNTIME"), "selector")] = w.Name
```

Either that, or we can just use the existing `app.kubernetes.io/name` as you suggested.
What do you think?

TomerFi: Implemented the changes request here: https://github.com/wildfly/wildfly-operator/pull/210/commits/94690ba53d1a12af15706f152a3bdf57de3a0fc9

The existing `app.kubernetes.io/name` label, valued with the unique CR name,
is now also being used for the HPA as the selector label.

jmesnil (approved):
Looks good to me, thanks!

jmesnil:
This fixes #208 & #208

BobVanB:
Good evening, any progress with this pull request?

jmesnil:
I'm closing your PR as it has been merged in the aggregated PR #215. Thanks!

PR #212, opened 2022-01-04
Add support for resources on the Stateful Set

BobVanB:
**Feature request:**

Add support for resources on `Container` in the `Stateful Set`

* Replaced some Unicode characters.
* Update the api version references.

**Credits**

Copied the test from #210 to see if its works and it did.
#210 fix: fixed autoscaling with the hpa

**Need help with:**

* Comments
* Docs

**This fixes:**

We use a open policy agent to enforce limits and requests on resources.
Beside this resource problem, we didn't see any problem with it.

**What i didn't do**

* `make codegen` to minimize the changes and i don't know if i should.
* validation, assuming that the error of the api of kubernetes is thrown.

### DIFF
--- a/deploy/crds/quickstart-resources-cr.yaml
+++ b/deploy/crds/quickstart-resources-cr.yaml
@@ -1,0 +1,16 @@
+---
+apiVersion: wildfly.org/v1alpha1
+kind: WildFlyServer
+metadata:
+  name: quickstart
+spec:
+  applicationImage: "quay.io/wildfly-quickstarts/wildfly-operator-quickstart:18.0"
+  replicas: 2
+  resources:
+    limits:
+      cpu: 500m
+      memory: 2Gi
+    requests:
+      cpu: 100m
+      memory: 1Gi
+...
--- a/doc/apis.adoc
+++ b/doc/apis.adoc
@@ -11,7 +11,7 @@ This document describes the types introduced by the WildFly Operator to be consu
 [options="header,footer"]
 |=======================
 | Field  | Description |Scheme| Required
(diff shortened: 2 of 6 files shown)


BobVanB:
@jmesnil i think you mean #210
The main difference between that one and mine is the implementation of the resources.

```golang
//  pkg/resources/statefulsets/statefulset.go 
if w.Spec.Resources != nil {
	statefulSet.Spec.Template.Spec.Containers[0].Resources = *w.Spec.Resources
}
```

My pull request also address this comment:
https://github.com/wildfly/wildfly-operator/blob/master/pkg/resources/statefulsets/statefulset.go#L153
```golang
// Kubernetes initializes empty map as nil, the reflect.DeepEqual causes troubles when we do not use nil too
if pvcTemplate.Spec.Resources.Limits != nil && len(pvcTemplate.Spec.Resources.Limits) == 0 {
	pvcTemplate.Spec.Resources.Limits = nil
}
if pvcTemplate.Spec.Resources.Requests != nil && len(pvcTemplate.Spec.Resources.Requests) == 0 {
	pvcTemplate.Spec.Resources.Requests = nil
}
```

But both pull request adds resources, so i don't mind.
I can rebase it if needed.

jmesnil:
@BobVanB thanks. Your PR is also aligning the resource creation for the statefulset and the pvc which is great.

Once #210 is merged, I'll ask you to rebase yours on top and we should be good.
thanks!

jmesnil:
I'm closing your PR as it has been merged in the aggregated PR #215. Thanks!

PR #227, opened 2022-06-17
[226] Use patch instead of update to change pod labels

### LINKED ISSUE #226: OCP 4.11 with EAP operator can not scale down to 0
When the Operator is used by a non-admin user on OCP 4.11, the following message is logged:

```
Forbidden: pod updates may not change fields other than `spec.containers[*].image`, `spec.initContainers[*].image`, `spec.activeDeadlineSeconds`, `spec.tolerations` (only additions to existing tolerations) or `spec.terminationGracePeriodSeconds` (allow it to be set to 1 if it was previously negative)
```

This error prevents scaling down the pods.

yersan:
fixes #226

### DIFF
--- a/pkg/resources/actions.go
+++ b/pkg/resources/actions.go
@@ -80,6 +80,25 @@ func Update(w *wildflyv1alpha1.WildFlyServer, client client.Client, objectDefini
 	return nil
 }
 
+func Patch(w *wildflyv1alpha1.WildFlyServer, client client.Client, objectDefinition runtime.Object, patch client.Patch) error {
+	logger := logWithValues(w, objectDefinition)
+	logger.Info("Patching Resource")
+
+	meta := objectDefinition.(metav1.Object)
+	// mark the object with the current wildflyserver generation unless it is the wildflyserver itself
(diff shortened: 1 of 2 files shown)


pkg/resources/actions.go
```diff
@@ -80,6 +80,25 @@ func Update(w *wildflyv1alpha1.WildFlyServer, client client.Client, objectDefini
...
+	if objectDefinition != w {
+		MarkServerGeneration(w, meta)
+	}
+
+	if err := client.Patch(context.TODO(), objectDefinition, patch); err != nil {
+		logger.Error(err, "Failed to patch resource")
+		return err
+	}
+
+	logger.Info("Patched resource")
```
jmesnil, line 98: @yersan should this be logged at `Debug` level?

yersan: @jmesnil I've changed the message to debug level for the trace that confirms the action taken. There is no need to show it as info, if the operation fails an error is logged, otherwise, we can assume it was executed successfully, no need to log such a confimation

PR #229, opened 2022-07-15
[226] Use patch instead of update to change pod status during TX reco…

### LINKED ISSUE #226: OCP 4.11 with EAP operator can not scale down to 0

yersan:
…very

Fixes #226

### DIFF
--- a/pkg/controller/wildflyserver/transaction_recovery.go
+++ b/pkg/controller/wildflyserver/transaction_recovery.go
@@ -80,8 +80,9 @@ func (r *ReconcileWildFlyServer) checkRecovery(reqLogger logr.Logger, scaleDownP
 		// The pod was already searched for the recovery port, marking that into the annotations
 		annotations := wfly.MapMerge(
 			scaleDownPod.GetAnnotations(), map[string]string{markerRecoveryPort: strconv.FormatInt(int64(scaleDownPodRecoveryPort), 10)})
+		patch := client.MergeFrom(scaleDownPod.DeepCopyObject())
 		scaleDownPod.SetAnnotations(annotations)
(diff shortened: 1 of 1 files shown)


pkg/controller/wildflyserver/transaction_recovery.go
```diff
@@ -80,8 +80,9 @@ func (r *ReconcileWildFlyServer) checkRecovery(reqLogger logr.Logger, scaleDownP
 		// The pod was already searched for the recovery port, marking that into the annotations
 		annotations := wfly.MapMerge(
 			scaleDownPod.GetAnnotations(), map[string]string{markerRecoveryPort: strconv.FormatInt(int64(scaleDownPodRecoveryPort), 10)})
+		patch := client.MergeFrom(scaleDownPod.DeepCopyObject())
```
mchoma, line 83: For me to understand. This line is to get origin value of `scaleDownPod`. Then we change `scaleDownPod` to desired value and pass both values to Patch method.

Maybe I would use different naming of `patch variable`, but I see it is aligned with previous reviewed fix [1] So probably it is OK.

[1] https://github.com/wildfly/wildfly-operator/pull/227/files

mchoma (approved):

PR #235, opened 2022-10-18
[234] Add the ability to configure the Readiness, Liveness and Startup probes

### LINKED ISSUE #234: Add possibility to customize readiness and liveness probe settings
According to https://access.redhat.com/documentation/en-us/red_hat_jboss_enterprise_application_platform/7.2/html/getting_started_with_jboss_eap_for_openshift_online/migrating-application-openshift-4 
Probes should be
`livenessProbe:
        exec:
          command:
            - /bin/bash
            - '-c'
            - /opt/eap/bin/livenessProbe.sh
        initialDelaySeconds: 60
        periodSeconds: 16
        successThreshold: 1
        failureThreshold: 3`
but but default they are still 10 seconds for statefulset. Please change their default values or add capability to modify them

yersan:
Fixes #234


config/crd/bases/wildfly.org_wildflyservers.yaml
```diff
@@ -209,6 +209,73 @@ spec:
                 minItems: 1
                 type: array
                 x-kubernetes-list-type: atomic
+              livenessProbe:
+                description: LivenessProbe defines the periodic probe of container
+                  liveness. Container will be restarted if the probe fails
```
jbliznak, line 214: minor, all other descriptions not ending with url have trailing `.`

api/v1alpha1/wildflyserver_types.go
```diff
@@ -67,6 +67,40 @@ type WildFlyServerSpec struct {
...
+	// How often (in seconds) to perform the probe.
+	// Default to 10 seconds. Minimum value is 1.
+	// +optional
+	PeriodSeconds int32 `json:"periodSeconds,omitempty"`
+	// Minimum consecutive successes for the probe to be considered successful after having failed.
+	// Defaults to 1. Must be 1 for liveness and startup. Minimum value is 1.
+	// +optional
+	SuccessThreshold int32 `json:"successThreshold,omitempty"`
+	// Minimum consecutive failures for the probe to be considered failed after having succeeded.
+	// Defaults to 3. Minimum value is 1.
```
jbliznak, line 101: where is eg. this default declared?

controllers/wildflyserver_controller_test.go
```diff
@@ -494,3 +495,310 @@ func TestWildFlyServerWithResources(t *testing.T) {
...
+	assert.Equal(replicas, *statefulSet.Spec.Replicas)
+	assert.Equal(applicationImage, statefulSet.Spec.Template.Spec.Containers[0].Image)
+
+	// check the Probes values
+	stsLivenessProbe := statefulSet.Spec.Template.Spec.Containers[0].LivenessProbe
+	assert.Equal(int32(60), stsLivenessProbe.InitialDelaySeconds)
+	assert.Equal(int32(0), stsLivenessProbe.TimeoutSeconds)
+	assert.Equal(int32(0), stsLivenessProbe.PeriodSeconds)
+	assert.Equal(int32(0), stsLivenessProbe.SuccessThreshold)
+	assert.Equal(int32(0), stsLivenessProbe.FailureThreshold)
```
jbliznak, line 705: > // Defaults to 3. Minimum value is 1.

how this can then be equal 0?

api/v1alpha1/wildflyserver_types.go
```diff
@@ -67,6 +67,40 @@ type WildFlyServerSpec struct {
...
+	StartupProbe *ProbeSpec `json:"startupProbe,omitempty"`
+}
+
+// ProbeSpec Probe describes a health check to be performed against a container to determine whether it is alive or ready to receive traffic.
+// +k8s:openapi-gen=true
+type ProbeSpec struct {
+	// Number of seconds after the container has started before liveness probes are initiated.
+	// Defaults to 60 seconds. Minimum value is 0.
+	// +optional
+	InitialDelaySeconds int32 `json:"initialDelaySeconds,omitempty"`
```
jmesnil, line 86: it is missing a `kubebuilder` instruction to exert the constraints:

`// +kubebuilder:validation:Minimum=0`

api/v1alpha1/wildflyserver_types.go
```diff
@@ -67,6 +67,40 @@ type WildFlyServerSpec struct {
...
+	// Number of seconds after which the probe times out.
+	// Defaults to 1 second. Minimum value is 1.
+	// More info: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle#container-probes
+	// +optional
+	TimeoutSeconds int32 `json:"timeoutSeconds,omitempty"`
```
jmesnil, line 91: `// +kubebuilder:validation:Minimum=1`

api/v1alpha1/wildflyserver_types.go
```diff
@@ -67,6 +67,40 @@ type WildFlyServerSpec struct {
...
+	// How often (in seconds) to perform the probe.
+	// Default to 10 seconds. Minimum value is 1.
+	// +optional
+	PeriodSeconds int32 `json:"periodSeconds,omitempty"`
```
jmesnil, line 95: `// +kubebuilder:validation:Minimum=1`

api/v1alpha1/wildflyserver_types.go
```diff
@@ -67,6 +67,40 @@ type WildFlyServerSpec struct {
...
+	// Minimum consecutive successes for the probe to be considered successful after having failed.
+	// Defaults to 1. Must be 1 for liveness and startup. Minimum value is 1.
+	// +optional
+	SuccessThreshold int32 `json:"successThreshold,omitempty"`
```
jmesnil, line 99: `// +kubebuilder:validation:Minimum=1`

api/v1alpha1/wildflyserver_types.go
jmesnil: `// +kubebuilder:validation:Minimum=1`

api/v1alpha1/wildflyserver_types.go
```diff
@@ -67,6 +67,40 @@ type WildFlyServerSpec struct {
...
+	LivenessProbe *ProbeSpec `json:"livenessProbe,omitempty"`
+	// ReadinessProbe defines the periodic probe of container service readiness. Container will be removed from service endpoints if the probe fails.
+	ReadinessProbe *ProbeSpec `json:"readinessProbe,omitempty"`
+	// StartupProbe indicates that the Pod has successfully initialized. If specified, no other probes are executed until this completes successfully.
+	// If this probe fails, the Pod will be restarted, just as if the livenessProbe failed. This can be used to provide different probe parameters at the beginning of a Pod's lifecycle,
+	// when it might take a long time to load data or warm a cache, than during steady-state operation.
+	StartupProbe *ProbeSpec `json:"startupProbe,omitempty"`
+}
+
+// ProbeSpec Probe describes a health check to be performed against a container to determine whether it is alive or ready to receive traffic.
```
jmesnil, line 80: It'd be worth mentioning that the operator will configure the exec/httpGet fields and only opens up configuration of the various timeouts and thresholds

jmesnil (changes requested):

api/v1alpha1/wildflyserver_types.go
yersan: Hi @jbliznak , about the defaults:

- The default values are added by the Kubernetes API; except for the `initialDelayInSeconds`. The Operator changes these defaults to:
1. Liveness Probe: 60
2. Readiness Probe: 10
3. Startup Probe: Use the Kibernetes defaults, which is 0.
I need to adjust the documentation to reflect it.

controllers/wildflyserver_controller_test.go
jbliznak: So to wrap it up as only `InitialDelaySeconds` is allowed to have 0 a we redefine default for that one it is safe for other properties to be initialized with 0 with meaning as uninitialized - ok then.

yersan: Yes, right.

config/crd/bases/wildfly.org_wildflyservers.yaml
yersan: I've added the missing trailing `.`

yersan: But I forgot to generate the CDR :(, I'll do it now

api/v1alpha1/wildflyserver_types.go
yersan: I've added the missing `// +kubebuilder:validation:Minimum=1`

api/v1alpha1/wildflyserver_types.go
yersan: I've added the missing `// +kubebuilder:validation:Minimum=0`

api/v1alpha1/wildflyserver_types.go
yersan: I've added the missing `// +kubebuilder:validation:Minimum=1`

api/v1alpha1/wildflyserver_types.go
yersan: I've added the missing` // +kubebuilder:validation:Minimum=1`

api/v1alpha1/wildflyserver_types.go
yersan: Thanks for the review @jmesnil , I've updated the description accordingly

api/v1alpha1/wildflyserver_types.go
```diff
@@ -67,6 +67,47 @@ type WildFlyServerSpec struct {
...
+	StartupProbe *ProbeSpec `json:"startupProbe,omitempty"`
+}
+
+// ProbeSpec describes a health check to be performed against a container to determine whether it is alive or ready to receive traffic.
+// The Operator will configure the exec/httpGet fields. Notice these fields are not exposed to the user since they are an implementation detail
+// that depends on the server image used to create the application image.
+// +k8s:openapi-gen=true
+type ProbeSpec struct {
+	// Number of seconds after the container has started before liveness probes are initiated.
+	// It defaults to 60 seconds for liveness probe. It defaults to 10 seconds for readiness probe. it defaults to 0 for startup probe.
```
jbliznak, line 86: > it defaults to 0 for startup probe.

First character should be capitalized

api/v1alpha1/wildflyserver_types.go
```diff
@@ -67,6 +67,47 @@ type WildFlyServerSpec struct {
...
+	// when it might take a long time to load data or warm a cache, than during steady-state operation.
+	StartupProbe *ProbeSpec `json:"startupProbe,omitempty"`
+}
+
+// ProbeSpec describes a health check to be performed against a container to determine whether it is alive or ready to receive traffic.
+// The Operator will configure the exec/httpGet fields. Notice these fields are not exposed to the user since they are an implementation detail
+// that depends on the server image used to create the application image.
+// +k8s:openapi-gen=true
+type ProbeSpec struct {
+	// Number of seconds after the container has started before liveness probes are initiated.
```
jbliznak, line 85: we could leave out the `liveness` word here as the sentence is common for all type of probes, right?

api/v1alpha1/wildflyserver_types.go
yersan: Absolutely right @jbliznak , sorry for the obvious errors

pkg/resources/statefulsets/statefulset.go
```diff
@@ -276,56 +279,148 @@ func createResources(r *corev1.ResourceRequirements) corev1.ResourceRequirements
...
+
+	return nil
+}
+
+// creates the common parts for the Linevess and Startup Probes
+func createLivenessStartupCommonProbe(scriptDefined bool, bootableJar bool, livenessProbeScript string, initialDelaySeconds int32, timeoutSeconds int32, periodSeconds int32, successThreshold int32, failureThreshold int32) *corev1.Probe {
+	if scriptDefined && !bootableJar {
+		livenessProbeScriptFallback, definedFallback := os.LookupEnv("SERVER_LIVENESS_SCRIPT_FALLBACK")
+		if !definedFallback {
+			livenessProbeScriptFallback = "curl 127.0.0.1:" + strconv.Itoa(int(resources.HTTPManagementPort)) + "/health/live"
```
jbliznak, line 396: This requires that the management API is bound to 0.0.0.0 or to loopback directly, right?

pkg/resources/statefulsets/statefulset.go
```diff
@@ -276,56 +279,148 @@ func createResources(r *corev1.ResourceRequirements) corev1.ResourceRequirements
...
 // *and* the application is not using Bootable Jar.
+// Otherwise, it creates a HTTPGet probe that checks the /health/live endpoint on the admin port.
 //
-// If defined, the SERVER_READINESS_SCRIPT env var must be the path of a shell script that
+// If defined, the SERVER_LIVENESS_SCRIPT env var must be the path of a shell script that
 // complies to the Kubernetes probes requirements.
-func createReadinessProbe(w *wildflyv1alpha1.WildFlyServer) *corev1.Probe {
-	readinessProbeScript, defined := os.LookupEnv("SERVER_READINESS_SCRIPT")
-	if defined && !w.Spec.BootableJar {
+// If SERVER_LIVENESS_SCRIPT script does not exist, then the Probe will execute the script defined by SERVER_LIVENESS_SCRIPT_FALLBACK
```
jbliznak, line 366: the same type of comment should also be there for createReadinessProbe

pkg/resources/statefulsets/statefulset.go
```diff
@@ -276,56 +279,148 @@ func createResources(r *corev1.ResourceRequirements) corev1.ResourceRequirements
...
+// If this script does not exist, the probe will execute and http get by using curl.
+func createStartupProbe(w *wildflyv1alpha1.WildFlyServer) *corev1.Probe {
+	livenessProbeScript, defined := os.LookupEnv("SERVER_LIVENESS_SCRIPT")
+
+	var initialDelaySeconds int32 = 60
```
jbliznak, line 371: api/v1alpha1/wildflyserver_types.go line 86 says

> It defaults to 60 seconds for liveness probe. It defaults to 10 seconds for readiness probe. It defaults to 0 seconds for startup probe.

controllers/wildflyserver_controller_test.go
```diff
@@ -494,3 +495,310 @@ func TestWildFlyServerWithResources(t *testing.T) {
...
+
+	// Check if stateful set has been created and has the correct size.
+	statefulSet := &appsv1.StatefulSet{}
+	err = cl.Get(context.TODO(), req.NamespacedName, statefulSet)
+	require.NoError(t, err)
+	assert.Equal(replicas, *statefulSet.Spec.Replicas)
+	assert.Equal(applicationImage, statefulSet.Spec.Template.Spec.Containers[0].Image)
+
+	// check the Probes values
+	stsLivenessProbe := statefulSet.Spec.Template.Spec.Containers[0].LivenessProbe
```
jbliznak, line 555: it's obvious from the impl method but could add this (I don't know the correct Go syntax :) but you get the idea) to check the default behaviour completely?

```
stsStartupProbe := statefulSet.Spec.Template.Spec.Containers[0].StartupProbe

assert stsStartupProbe is nil 
```

api/v1alpha1/wildflyserver_types.go
jbliznak: missed this one, ` // +kubebuilder:validation:Minimum=1` is still missing for `FailureThreshold`

pkg/resources/statefulsets/statefulset.go
yersan: Yes, that's correct, if the management interface is not bound to 0.0.0.0 or to loopback, the probe will fail

pkg/resources/statefulsets/statefulset.go
```diff
@@ -276,56 +279,148 @@ func createResources(r *corev1.ResourceRequirements) corev1.ResourceRequirements
...
+}
+
+// creates the common parts for the Linevess and Startup Probes
+func createLivenessStartupCommonProbe(scriptDefined bool, bootableJar bool, livenessProbeScript string, initialDelaySeconds int32, timeoutSeconds int32, periodSeconds int32, successThreshold int32, failureThreshold int32) *corev1.Probe {
+	if scriptDefined && !bootableJar {
+		livenessProbeScriptFallback, definedFallback := os.LookupEnv("SERVER_LIVENESS_SCRIPT_FALLBACK")
+		if !definedFallback {
+			livenessProbeScriptFallback = "curl 127.0.0.1:" + strconv.Itoa(int(resources.HTTPManagementPort)) + "/health/live"
+		}
+		probeScript := "if [ -f " + livenessProbeScript + " ]; then " + livenessProbeScript + "; else " + livenessProbeScriptFallback + "; fi"
```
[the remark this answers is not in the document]
yersan: livenessProbeScriptFallback should be controlled internally, it is something that is not expected to be modified by the users. It just adds us the possibility to tweak the script under certain circumstances.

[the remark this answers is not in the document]
jbliznak: yeah, but there could still be a possibility the file is not available (eg. the container filesystem is not available at the moment) - and in that case this executing inaccessible file would end up in the probe error which in turn means "just ignore the error and make it look like the probe succeeded" on older versions of OpenShift/Kubernetes - as the issue stated

api/v1alpha1/wildflyserver_types.go
yersan: Correct, fixed. I've just added // +kubebuilder:validation:Minimum=1 under `FailureThreshold`

pkg/resources/statefulsets/statefulset.go
yersan: Fixed.

pkg/resources/statefulsets/statefulset.go
yersan: Fixed.

jbliznak (approved):
Thanks Yeray for fixing my concern, I have no more objections.

jmesnil (approved):

pkg/resources/statefulsets/statefulset.go
```diff
@@ -276,56 +279,148 @@ func createResources(r *corev1.ResourceRequirements) corev1.ResourceRequirements
...
+	if w.Spec.ReadinessProbe != nil {
+		initialDelaySeconds = w.Spec.ReadinessProbe.InitialDelaySeconds
+		timeoutSeconds = w.Spec.ReadinessProbe.TimeoutSeconds
+		periodSeconds = w.Spec.ReadinessProbe.PeriodSeconds
+		successThreshold = w.Spec.ReadinessProbe.SuccessThreshold
+		failureThreshold = w.Spec.ReadinessProbe.FailureThreshold
+	}
+
 	if defined && !w.Spec.BootableJar {
+		readinessProbeScriptFallback, definedFallback := os.LookupEnv("SERVER_READINESS_SCRIPT_FALLBACK")
```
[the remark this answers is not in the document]
yersan: It just opens to us the possibility to tweak the script to work around if we need it. We should not provide a value for these variables

doc/apis.adoc
```diff
@@ -40,20 +40,38 @@ It uses a `StatefulSet` with a pod spec that mounts the volume specified by `sto
...
-| `configMaps` | List of ConfigMap names to mount as volumes in the containers. Each config map is mounted as a read-only volume under `/etc/configmaps/<config map name>` | string[] | false
-| `disableHTTPRoute`| Disable the creation a route to the HTTP port of the application service (false if omitted) | bool | false
 | `sessionAffinity`| If connections from the same client IP are passed to the same WildFlyServer instance/pod each time (false if omitted) | bool | false
+| `standaloneConfigMap` | spec to specify how standalone configuration can be read from a `ConfigMap` | *<<standaloneconfigmapspec>> |false
+| `startupProbe` | Indicates that the Pod has successfully initialized. If specified, no other probes are executed until this completes successfully. If this probe fails, the Pod will be restarted, just as if the livenessProbe failed. This can be used to provide different probe parameters at the beginning of a Pod's lifecycle, when it might take a long time to load data or warm a cache, than during steady-state operation. | *<<probespec>> |false
```
jmesnil, line 58: @yersan It could be worth mentioning that, contrary to the `livenessProbe` and `readinessProbe`, the `startupProbe` will not be created by default unless the user explicitly configure it in the spec.

jmesnil (changes requested):

doc/apis.adoc
yersan: Done.

jmesnil (approved):

api/v1alpha1/wildflyserver_types.go
```diff
@@ -69,6 +69,54 @@ type WildFlyServerSpec struct {
...
+	// Minimum consecutive failures for the probe to be considered failed after having succeeded.
+	// Defaults to 3. Minimum value is 1.
+	// +kubebuilder:validation:Minimum=1
+	// +optional
+	FailureThreshold int32 `json:"failureThreshold,omitempty"`
+	// Depending on the kind of base application image, the Operator will configure an HTTP or Exec probe. For example, to keep compatibility mode with legacy servers, all non-Bootable JAR based application images
+	// will be configured to use Exec probes by default. This flag allows you to force switching the probe configuration to use an HTTP request instead of executing a script. Notice this flag works in one way only, configuring http:false
+	// does not imply the probe will be forced to switch to an Exec probe instead of HTTP one.
+	// +optional
+	// +kubebuilder:default:=false
```
jmesnil, line 118: @yersan specifying a default with `+kubebuilder` will explicitly set the field to `false` when a CR is created:

```yaml
kind: WildFlyServer
metadata:
  name: example
  namespace: jmesnil-dev
spec:
  readinessProbe:
    http: false
  livenessProbe:
    http: false
  startupProbe:
    http: false
  replicas: 0
```

Instead, I think we should remove the `+kubebuilder:default` annotation and do the same thing we are doing for `bootableJar`:

```
// If omitted, it defaults to false (startup probe uses an Exec probe)
```

jmesnil (changes requested):

api/v1alpha1/wildflyserver_types.go
yersan: Done. I've removed the default CRD value

jmesnil (approved):

controllers/wildflyserver_controller_test.go
```diff
@@ -580,3 +581,393 @@ func TestWildFlyServerWithSecurityContext(t *testing.T) {
 	assert.Equal(readOnlyRootFilesystem, statefulSet.Spec.Template.Spec.Containers[0].SecurityContext.ReadOnlyRootFilesystem)
 	assert.Equal(runAsNonRoot, statefulSet.Spec.Template.Spec.Containers[0].SecurityContext.RunAsNonRoot)
 }
+
+func TestWildFlyServerWithHttpProbes(t *testing.T) {
```
tommaso-borgato, line 585: @yersan I don't understand what triggers the usage of HTTP probes here; IIUC default (no `bootableJar` or `http` setting) should be exec probes

tommaso-borgato: @yersan shouldn't we expect exec probes like e.g. ```        exec:
          command:
            - /bin/bash
            - '-c'
            - >-
              if [ -f '/opt/eap/bin/livenessProbe.sh' ]; then
              /opt/eap/bin/livenessProbe.sh; else curl --fail
              http://127.0.0.1:9990/health/live; fi``` in this case?

yersan: Hi @tommaso-borgato , the intention of this test is to verify the creation of probes that use pure HTTP request, not the Exec commands. I used the presence of the aforementioned environment variables to force the switch that triggers it. 

Would it get clearer if we use bootableJar=true? That would be more close to the reality.

The verification of the Exec probes is done in the TestWildFlyServerWithProbesScript

yersan: > which hasn't either bootableJar or http set, when processed by the operator produces HTTP Probes instead of exec probes;

Because the env variables SERVER_LIVENESS_SCRIPT and SERVER_READINESS_SCRIPT are not configured for this test. If they are not configured (for example the upstream operator does not configure them), then we configure the probes using HTTP get requests.

tommaso-borgato: @yersan Thank you for the explanation, now I see it

yersan:
@tommaso-borgato This is now ready for your review accordingly whit what we have been discussing by removing the HTTP field and exposing the Exec and HTTPGet

controllers/wildflyserver_controller_test.go
```diff
@@ -580,3 +583,711 @@ func TestWildFlyServerWithSecurityContext(t *testing.T) {
...
+	stsLivenessProbe := statefulSet.Spec.Template.Spec.Containers[0].LivenessProbe
+	assert.Equal(int32(60), stsLivenessProbe.InitialDelaySeconds)
+	assert.Equal(int32(0), stsLivenessProbe.TimeoutSeconds)
+	assert.Equal(int32(0), stsLivenessProbe.PeriodSeconds)
+	assert.Equal(int32(0), stsLivenessProbe.SuccessThreshold)
+	assert.Equal(int32(0), stsLivenessProbe.FailureThreshold)
+	assert.True(reflect.DeepEqual(livenessExec.Command, stsLivenessProbe.Exec.Command))
+	assert.Nil(stsLivenessProbe.HTTPGet)
+}
+
```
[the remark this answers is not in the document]
yersan: Hello @tommaso-borgato I don't see it necessary since there is no interaction between those conditions in the source code. Each probe uses its own method, and the existence of one vs another does not affect to any logic.

We always invoke each of the methods See https://github.com/wildfly/wildfly-operator/blob/a4764d07a3afe3541d24cd3aa2367117f765bbcf/pkg/resources/statefulsets/statefulset.go#L110-L114  

Anyway, since tests are quick and the change is easy, we can add it, but it is unnecessary from my point of view.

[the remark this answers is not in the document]
tommaso-borgato: ok, no need to if they are totally independent code paths

tommaso-borgato:
@yersan I see the `bundle/manifests/wildfly.org_wildflyservers.yaml` file is not updated

yersan:
> @yersan I see the `bundle/manifests/wildfly.org_wildflyservers.yaml` file is not updated

@tommaso-borgato We don't need to update it. It is confusing, so I guess the best would be to remove it from Git. This is the CRD generated when you are creating a bundle for the Operator. It has nothing to do with this PR.

PR #255, opened 2023-01-31
Update GitHub Actions

### LINKED ISSUE #254: Setup minikube fails with the following error: Unsupported OS, action only works in Ubuntu 18 or 20
The action used to setup minikube only supports Ubuntu 18 or 20, we should update the Github Action and use a valid Ubuntu version

yersan:
Fixes #254

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -14,17 +14,17 @@ on:
 jobs:
 
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
     - name: Set up Go
-      uses: actions/setup-go@v2
+      uses: actions/setup-go@v3
       with:
         go-version: 1.17.7
     - name: Build
       run: make build
     - name: Set up Minikube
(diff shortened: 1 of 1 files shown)


yersan:
The image used for the clustering test case is using cluster-bench 10 instead of 9 with a widlfly-s2i image based on WF26, so the test fails.

@jfdenise , can you please update the cluster bench image available on quay.io to use cluster bench v3? We need to fix this image in any case.

```
s2i build https://github.com/clusterbench/clusterbench.git \
          -e MAVEN_ARGS_APPEND=-Pee8 \
          -e S2I_SOURCE_DEPLOYMENTS_DIR=clusterbench-ee8-ear/target \
          -e GALLEON_PROVISION_LAYERS=cloud-server,web-clustering,ejb,-ejb-local-cache,ejb-dist-cache,jsf \
          -r 3.x \
          quay.io/wildfly/wildfly-centos7:26.1.3 \
          wildfly-s2i-builder-image

docker tag wildfly-s2i-builder-image:latest quay.io/wildfly-quickstart/clusterbench:latest
docker push quay.io/wildfly-quickstart/clusterbench:latest
```

In any case, we have to change improve the uses of these examples and use a quickstart for clustering and allow manual build of the example images via Helm Charts.

PR #238, opened 2022-11-03
Add security context

danielpalstra:
Add securityContext to the WildflyServer spec so that the STS can spawn Pods with limited privileges.

PR #267, opened 2023-03-13
[#266] [0.5.7] Update Pods Security Standards

### LINKED ISSUE #266: Update Pod Security Standards
Kubernetes Pod Security Standards are being updated (as explained in https://sdk.operatorframework.io/docs/best-practices/pod-security-standards/) and the WildFly Operator lags behind these changes.

In the `main`branch, the user can specify the `SecurityContext` on the deployed Pod templates to control the behaviour (#238).

However, it is missing a `seccompProfile` to comply with the security standards.

In addition, we can provide a good default value for this security context and do not need to ask the user to specify it.

jmesnil:
Configure the statefulset's pods to run in non-root mode. It drops all security capabilities and does not allow priviledge escalation.

This fixes #266 WIP add security context to statefulset

### DIFF
--- a/pkg/resources/statefulsets/statefulset.go
+++ b/pkg/resources/statefulsets/statefulset.go
@@ -56,6 +56,12 @@ func NewStatefulSet(w *wildflyv1alpha1.WildFlyServer, labels map[string]string, 
 		wildflyImageTypeAnnotation = resources.ImageTypeBootable
 	}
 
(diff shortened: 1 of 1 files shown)


jmesnil:
In `main` branch, the `SecurityContext` field is exposed to the API and does not provide a default value #238.

I'll open a follow up PR against `main` to use the same default value

jmesnil:
PR for `main` is #268

PR #269, opened 2023-03-15
Load wildfly application images in Minikube

jmesnil:
to make the tests run faster as they will not have to pull the images


jmesnil:
Compare to the previous CI job on `main`, the CI tests went from [`15m14`](https://github.com/wildfly/wildfly-operator/actions/runs/4414374231/jobs/7735996392) to [11m25](https://github.com/wildfly/wildfly-operator/actions/runs/4426883834/jobs/7763724005).

That's a 4-minutes improvement and this makes the test timed out less likely

yersan (approved):

PR #271, opened 2023-04-12
[#266] Updated Pod security standards

jmesnil:
* Do not specify the 185 (jboss) user by default


yersan (approved):
Omitting the creation of a default context will affect also to the SecurityContext documentation, so we should change the documentation accordingly.

